PR #1, opened 2023-05-12
Add Carmen in elementary form

hkalina:
Limitation:
* only basic run (fakenet 1/1 without rpc request) tested for now
* the live StateDB is always used - this needs to be changed for TxPool (to the "latest" archive block) and for RPC (to the requested archive block)


carmenadapter/switch.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/ethereum/go-ethereum/core/state/snapshot"
+	"math/big"
+	"os"
+	"path/filepath"
+)
+
+var carmenState carmen.State
+var carmenStateDb carmen.StateDB
+
+func InitializeCarmen(impl string, datadir string) error {
```
HerbertJordan, line 18: Maybe we should use a more generalized naming schema, treating the DB variants equally and keeping the possibility for other implementations open, like it was done in Aida.

If so, I would propose to name this package just "state" or something similar and rename `InitializeCarmen` to `InitializeState`, tasked with initializing whatever implementation variant has been selected. And the rest of the project should be unaware what DB is actually used -- as is already almost archived by this PR.

HerbertJordan (changes requested):
Great work, thanks!

I have just a few minor requests and questions, please see the individual comments.

hkalina:
Updated:
* logs management in carmen
* mutex-protected parallel access into the live state - used for block processing, for txpool and also for RPC with block=latest

gossip/blockproc/evmmodule/evm.go
```diff
@@ -29,6 +29,10 @@ func (p *EVMModule) Start(block iblockproc.BlockCtx, statedb *state.StateDB, rea
 	if block.Idx != 0 {
 		prevBlockHash = reader.GetHeader(common.Hash{}, uint64(block.Idx-1)).Hash
 	}
+
+	// Start block (Carmen specific)
```
HerbertJordan, line 33: nit: the `(Carmen specific)` part should not be of concern for this function. Also below.

HerbertJordan (approved):

PR #119, opened 2024-04-12
Port Carmen integration to Carmen's new facade

HerbertJordan:
A draft on how the Carmen facade could be integrated into Sonic.


HerbertJordan:
Replaced by #144

PR #132, opened 2024-05-23
Avoid null topics in logs in RPC

thaarok:
This extends RPC api to never return a null Topics field in the getLogs / getTransactionReceipt responses.


HerbertJordan (approved):
Looks good, thanks!

PR #133, opened 2024-05-23
Verify that nil Topics in log does not propagate into genesis

thaarok:
Follow-up of #132 - prove that nil Topics are not issue of genesis files.


HerbertJordan (approved):
Thanks for verifying this!

PR #137, opened 2024-05-24
Add archive genesis Mainnet-282500

thaarok:


thaarok:
Duplicates https://github.com/Fantom-foundation/Sonic/pull/135

PR #141, opened 2024-05-29
Upgrade go-ethereum dependency to 1.14.3

thaarok:
This PR contains changes related to the go-ethereum upgrade.

Notes:
* SetExternalHash in EvmHeader.EthHeader is not necessary, as the method 

To be done in follow-ups:
* BlobHashes / BlobGasFeeCap / blobGasPrice should be handled (should not be used in current network configuration)
* GetStorageRoot should be wired to Carmen - depends on https://github.com/Fantom-foundation/Carmen/issues/887
* Check necessity of `ethHeader.SetExternalHash` usage in dummy_block.go - affects output of **eth_subscribe("newHeads")** / eth_newBlockFilter


thaarok:
As "Replace Header.SetExternalHash by custom type for subscribed headers" introduces code, which duplicates header serialization in the ethapi RPC package, refactoring PR (which moves both serializations into a single place) follows: https://github.com/Fantom-foundation/Sonic/pull/143

gossip/emitter/ordering.go
```diff
@@ -0,0 +1,179 @@
...
+// Clear removes the entire content of the heap.
+func (t *transactionsByPriceAndNonce) Clear() {
+	t.heads, t.txs = nil, nil
+}
+
+func (t *transactionsByPriceAndNonce) Copy() *transactionsByPriceAndNonce {
+	txsCopy := make(map[common.Address][]*txpool.LazyTransaction, len(t.txs))
+	for k, v := range t.txs {
+		txsCopy[k] = v
+	}
```
HerbertJordan, line 172: Could be simplified to
```suggestion
	txsCopy := maps.Clone(t.txs)
```

gossip/emitter/ordering.go
```diff
@@ -0,0 +1,179 @@
...
+	return &transactionsByPriceAndNonce{
+		txs:     txsCopy,
+		heads:   append(make(txByPriceAndTime, 0, t.heads.Len()), t.heads...),
```
HerbertJordan, line 175:
```suggestion
		heads:   slices.Clone(t.heads),
```

gossip/ethapi_backend.go
```diff
@@ -289,7 +289,10 @@ func (b *EthAPIBackend) GetReceiptsByNumber(ctx context.Context, number rpc.Bloc
 	}
 
 	block := b.state.GetBlock(common.Hash{}, uint64(number))
-	receipts := b.svc.store.evm.GetReceipts(idx.Block(number), b.signer, block.Hash, block.Transactions)
+	time := uint64(block.Time.Unix())
+	baseFee := block.BaseFee
+	blobGasPrice := new(big.Int) // TODO: fix a price for this
```
HerbertJordan, line 294: We have a range of lose ends regarding the blob pricing.

It would be great to have some systematic way to locate those when we work towards upgrading to Cancun. All of those will have to be addressed.

utils/rate/gauge.go
```diff
@@ -52,25 +52,25 @@ func (g *Gauge) rateToGauge(valuesSum, calls float64) float64 {
 // Rate1 returns the one-minute moving average of the gauge values.
 // Cannot be larger than max(largest input value, 0)
 func (g *Gauge) Rate1() float64 {
-	return g.rateToGauge(g.input.Rate1(), g.count.Rate1())
+	return g.rateToGauge(g.input.Snapshot().Rate1(), g.count.Snapshot().Rate1())
```
HerbertJordan, line 55: This function and the ones below should only take a single snapshot and retrieve the needed information from the same snapshot to be consistent.

HerbertJordan (approved):
This PR is way to extensive to be properly reviewed. Also, many of the changes are in areas I am not familiar with, so I can not judge its impact. I have to trust on your judgement on those. Maybe you want to involve additional reviewers?

I guess ultimately, we need to depend on extensive testing to make sure those changes do what they are intended to.

utils/rate/gauge.go
thaarok: `g.input` and `g.count` are standalone meters - I cannot obtain a single snapshot for both of them

PR #143, opened 2024-05-30
Refactor block/header serialization in RPC - remove duplicite code

thaarok:
The block/header JSON serialization is duplicated between ethapi (regular RPC queries) and filters (handling of `eth_subscribe(newHeads)`) - this moves this serialization into a single place and ensures the same format for both.

### DIFF
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -772,27 +772,33 @@ func (s *PublicBlockChainAPI) GetProof(ctx context.Context, address common.Addre
 func (s *PublicBlockChainAPI) GetHeaderByNumber(ctx context.Context, number rpc.BlockNumber) (*evmcore.EvmHeaderJson, error) {
 	header, err := s.b.HeaderByNumber(ctx, number)
 	if header != nil && err == nil {
-		response := s.rpcMarshalHeader(header, s.getBlockReceipts(ctx, number))
-		return response, err
+		receipts, err := s.getBlockReceipts(ctx, rpc.BlockNumber(header.Number.Uint64()))
+		if err != nil {
+			return nil, err
+		}
+		return s.rpcMarshalHeader(header, receipts), nil
 	}
 	return nil, err
 }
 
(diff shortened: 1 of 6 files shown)


ethapi/api.go
```diff
@@ -770,73 +769,51 @@ func (s *PublicBlockChainAPI) GetProof(ctx context.Context, address common.Addre
...
 	if block != nil && err == nil {
-		response, err := s.rpcMarshalBlock(block, s.calculateExtBlockApi(ctx, number), true, fullTx)
-		if err == nil && number == rpc.PendingBlockNumber {
-			// Pending blocks need to nil out a few fields
-			for _, field := range []string{"hash", "nonce", "miner"} {
-				response[field] = nil
-			}
-		}
-		return response, err
+		return s.rpcMarshalBlock(block, s.getBlockReceipts(ctx, number), true, fullTx)
```
HerbertJordan, line 806: In the old version there was a special case handling for pending blocks. Is this no longer needed?

This is also the case in other functions in this file.

ethapi/api.go
```diff
@@ -770,73 +769,51 @@ func (s *PublicBlockChainAPI) GetProof(ctx context.Context, address common.Addre
...
-			ext.bloom = types.CreateBloom(receipts)
-		} else {
-			ext.receiptsRoot = types.EmptyRootHash
-		}
+func (s *PublicBlockChainAPI) getBlockReceipts(ctx context.Context, blkNumber rpc.BlockNumber) types.Receipts {
+	if blkNumber == rpc.EarliestBlockNumber {
+		return nil
 	}
-	return ext
+	receipts, _ := s.b.GetReceiptsByNumber(ctx, blkNumber)
```
HerbertJordan, line 794: The error here should probably not be ignored.

HerbertJordan (approved):
As for other changes in the Sonic client, I can merely do a rough sanity check on the code. I can not judge the implication of those changes on the overall system.

PR #144, opened 2024-05-31
Port Carmen integration to Carmen's new facade

thaarok:
This replaces (and integrates) https://github.com/Fantom-foundation/Sonic/pull/119.

Integrates latest Carmen and the Carmen facade.


HerbertJordan (commented):
The changes look plausible, but this should be thoroughly tested. Do you have a list of tests to be conducted before merging this?

thaarok:
@HerbertJordan do you think we should make a more minimalist version of this PR?
I think to get a good test coverage of all this does not seems reasonable in the near future...

PR #170, opened 2024-07-10
Fix message to reference

jenikd:
Checking reference for message to variable and using recover for this tracetx function


thaarok:
For reference, this is fix of https://github.com/Fantom-foundation/go-opera/issues/468

magneticmeganlovesjhonny (commented):
Approved

PR #182, opened 2024-07-26
Fix parent block header reference

jenikd:
Fixing wrong reference of block header from parent to actual block which is replayed by debug_trace
Issue https://github.com/Fantom-foundation/Sonic/issues/180


thaarok (approved):
Oh, sorry, this was my very stupid fault

PR #188, opened 2024-08-02
Upgrade carmen to version with Amount type

thaarok:
From review in https://github.com/Fantom-foundation/Sonic/pull/187 it seems the Carmen upgrade and related migration from big.Int amounts to Carmen specific Amount type would be better to do in a standalone PR.

Because of the conversion, StateDB can now produce error which can be exported only through dedicated Error() method. Then we need to make sure this method is called after EVM run.


HerbertJordan (approved):
Looks good to me, thanks for factoring this out into its own PR.

PR #189, opened 2024-08-02
Add batch and js tracer limits

jenikd:
This PR adds possibility to limit size of rpc **batch** calls in general and number of concurent **javascript tracers** used in `debug_trace` rpc calls.

There are two new config variables
- BatchRequestLimit: 1000
- JSTracerLimit: 0

From command line, they can be used with parameters
- rpc.batchrequestlimit
- rpc.jstracerlimit

Default of `BatchRequestLimit` is set to value 1000 to stay compatible with newer version of go-ethereum, which also introduced this variable.
Default for `JSTracerLimit` is also 1000


gossip/config.go
```diff
@@ -219,6 +225,10 @@ func DefaultConfig(scale cachescale.Func) Config {
 		RPCGasCap:   50000000,
 		RPCTxFeeCap: 100, // 100 FTM
 		RPCTimeout:  5 * time.Second,
+
+		BatchRequestLimit: 1000,
+		// Limit 0 means, that there is no limit
+		JSTracerLimit: 0,
```
HerbertJordan, line 231: Should there be a limit by default?

ethapi/api.go
```diff
@@ -2177,9 +2178,11 @@ func (api *PublicDebugAPI) traceTx(ctx context.Context, message evmcore.Message,
...
 			timeout = rpcTimeout
 		}
-		if t, err := tracers.New(*config.Tracer, txctx); err != nil {
+		t, err := tracers.New(*config.Tracer, txctx)
+		if err != nil {
 			return nil, err
-		} else {
+		}
+		defer t.Destroy()
 			deadlineCtx, cancel := context.WithTimeout(ctx, timeout)
```
HerbertJordan, line 2186: nit: this file's code formatting seems to be broken

HerbertJordan (approved):
Looks good to me, but you should get another approval from somebody more familiar with the code base.

gossip/config.go
jenikd: After discussion with @jmpike changed limit to 1000

PR #216, opened 2024-10-03
Enable Cancun upgrade when Sonic upgrade is enabled

thaarok:
Enable Cancun upgrade when Sonic upgrade is enabled. Because Sonic is not enabled by default in the new networks, new fakenets are being created with Cancun already enabled.

The base Chainconfig structure "AllEthashProtocolChanges" is copied from Ethereum to avoid introducing EVM changes by go-ethereum dependency upgrades.

**This changes the database structure:**
Since the Cancun upgrade is enabled for the EVM by the block timestamp, the timestamp had to be added into the UpgradeHeights database structure. Appropriate database migration is included.
This also removes fantom mainnet specific hack "fixEventTxHashes".

@HerbertJordan I will welcome you review of the CarmenStateDB adapter and the EVM ChainConfig. (This part is short, but maybe you will have a better overview of consequences.)


kjezek (commented):
Changes related to `stateDB` adaptor to `Carmen` looks good to me.

kjezek (commented):
I noticed that this line:

https://github.com/Fantom-foundation/Sonic/blob/jk/upgrades/gossip/evmstore/carmen.go#L203

should call `CreateContract`, not `CreateAccount`

kjezek (commented):
this method body: https://github.com/Fantom-foundation/Sonic/blob/jk/upgrades/gossip/evmstore/carmen.go#L187

is not implemented, is it correct?

thaarok:
> should call `CreateContract`, not `CreateAccount`

Fixed.

> [SetStorage] is not implemented, is it correct?

Yes, SetStorage is not supported by Carmen. Its only usage is StateOverride in eth_call - the "state" field can be (for most of cases) easily replaced with "stateDiff", which is supported.

HerbertJordan (approved):
I am not familiar with the `UpdateHeight` mechanism nor the reasoning behind `fixEventTxHashes` (why it was needed and why it is dropped now). So I can not say anything really about this. The update to the Carmen adapter looks good though.

Any idea how we can test this?

thaarok:
Reported by @jenikd : this TODO needs to be resolved:
https://github.com/Fantom-foundation/Sonic/blob/2d21d0b56d2f4fc86a984531fcdfb094f210f47d/evmcore/state_transition.go#L270C2-L270C72

evmcore/evm.go
```diff
@@ -50,7 +50,7 @@ func NewEVMBlockContext(header *EvmHeader, chain DummyChain, author *common.Addr
 	if header.BaseFee != nil {
 		baseFee = new(big.Int).Set(header.BaseFee)
 	}
-	if header.Difficulty.Sign() == 0 {
+	if header.Difficulty != nil && header.Difficulty.Sign() == 0 {
```
thaarok, line 53: @jenikd Does header.Difficulty==nil happens? In which conditions? Are we sure here should not be `header.Difficulty == nil || header.Difficulty.Sign() == 0`?

jenikd: You are right, because I've added difficulty initialization into `ToEvmHeader` function, now in this place `header.Difficulty` will not be nil. Without it it happened, when I wanted to replay any block with debug.

evmcore/dummy_block.go
```diff
@@ -45,6 +45,9 @@ type (
 		GasUsed  uint64
 
 		BaseFee *big.Int
+
+		Difficulty  *big.Int
```
HerbertJordan, line 49: Since `Difficulty` is constant in Fantom (and presumably also in Sonic), I think this field is not required in our internal datastructures.

evmcore/dummy_block.go
```diff
@@ -45,6 +45,9 @@ type (
...
+		MixDigest   common.Hash
```
HerbertJordan, line 50: [EIP-4399](https://eips.ethereum.org/EIPS/eip-4399) refers to this field as `mixHash` and recommends its renaming to `prevRandao`. We should probably stick to the suggested naming for those fields.

evmcore/dummy_block.go
```diff
@@ -85,6 +94,8 @@ func ToEvmHeader(block *inter.Block, index idx.Block, prevHash hash.Event, rules
 		GasLimit:   math.MaxUint64,
 		GasUsed:    block.GasUsed,
 		BaseFee:    baseFee,
+		Difficulty: difficulty,
+		MixDigest:  common.Hash{}, // TODO provide pseudorandom data?
```
HerbertJordan, line 98: A plan to support this is in the making. However, it is beyond the scope of this PR.

HerbertJordan (changes requested):

evmcore/dummy_block.go
```diff
@@ -138,14 +149,13 @@ type EvmHeaderJson struct {
...
 	Bloom       types.Bloom      `json:"logsBloom"        gencodec:"required"`
-	Difficulty  *hexutil.Big     `json:"difficulty"       gencodec:"required"`
 	Number      *hexutil.Big     `json:"number"           gencodec:"required"`
 	GasLimit    hexutil.Uint64   `json:"gasLimit"         gencodec:"required"`
 	GasUsed     hexutil.Uint64   `json:"gasUsed"          gencodec:"required"`
 	Time        hexutil.Uint64   `json:"timestamp"        gencodec:"required"`
 	TimeNano    hexutil.Uint64   `json:"timestampNano"`
 	Extra       hexutil.Bytes    `json:"extraData"        gencodec:"required"`
-	MixDigest   common.Hash      `json:"mixHash"`
+	PrevRandao  common.Hash      `json:"prevRandao"`
```
HerbertJordan, line 158: If I read [EIP-4399](https://eips.ethereum.org/EIPS/eip-4399#block-structure) correctly, in the JSON file format the field is still to be named `mixHash`. Just internally, the renaming is suggested.

This Json format is the format we use for responding to RPC queries. In [docs](https://www.quicknode.com/docs/kaia/eth_getHeaderByNumber) this field is called `mixHash`.

evmcore/dummy_block.go
```diff
@@ -138,14 +149,13 @@ type EvmHeaderJson struct {
 	TxHash      common.Hash      `json:"transactionsRoot" gencodec:"required"`
 	ReceiptHash common.Hash      `json:"receiptsRoot"     gencodec:"required"`
 	Bloom       types.Bloom      `json:"logsBloom"        gencodec:"required"`
-	Difficulty  *hexutil.Big     `json:"difficulty"       gencodec:"required"`
 	Number      *hexutil.Big     `json:"number"           gencodec:"required"`
```
HerbertJordan, line 152: I think we need to preserve the difficulty field in the RPC output format, as, for instance, documented [here](https://www.quicknode.com/docs/kaia/eth_getHeaderByNumber).

HerbertJordan (changes requested):
The JSON output format should be aligned with the RPC spec. Other than that, it looks good to me.

HerbertJordan (dismissed):

HerbertJordan (approved):

PR #221, opened 2024-10-08
Update demo scripts for Sonic

thaarok:
This fixes scripts for running demo fakenet network.

Replaces #155.

### DIFF
--- a/demo/.gitignore
+++ b/demo/.gitignore
@@ -1,2 +1,3 @@
-opera*.datadir
+sonic*.datadir
+tool.datadir
 *.log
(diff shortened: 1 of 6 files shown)


jenikd (approved):

HerbertJordan (dismissed):

HerbertJordan:
Can this demo script be used as a smoke test for the client by running it as part of the CI or a nightly build?

jenikd:
> Can this demo script be used as a smoke test for the client by running it as part of the CI or a nightly build?

Yes it can be done. However I think, that Norma uses testnet for this kind of tests and not fakenet, so it is better. If there is no test with fakenet, that would be good to integfrate this.

HerbertJordan (dismissed):

jenikd (approved):

PR #223, opened 2024-10-09
Kjezek/enable ethereum test

kjezek:
This PR adapts ethereum tests to run against Sonic client.

It uses a customised version of ethereum tests in `go-ethereum-sonic` and passes carmen implementation into those test. 

It implements a factory and adaptor for carmen to be able to pass it to `go-ethereum-sonic`.

`init_test.go` was copied and slightly modified from original tests. It heavily relies on reflection to parse test data, unmarshall from json, and invokes tests functions via reflection.  I.e. it was decided to use it as a black box.  

TODO
- [x] go-ethereum-sonic deps must be updated for this PR to compile and work 
- [x] it needs Cancun related changes for tests to pass


gossip/evmstore/carmen.go
```diff
@@ -73,6 +73,27 @@ func (c *CarmenStateDB) GetLogs(txHash common.Hash, blockHash common.Hash) []*ty
 	return logs
 }
 
+func (c *CarmenStateDB) Logs() []*types.Log {
```
HerbertJordan, line 76: Is there any test that would verify that we produce correct logs and/or this conversion is correct?

If not, we should add this to the Todo list.

tests/init_test.go
HerbertJordan, line 1: This is still a lot of code that got duplicated and that somebody needs to know that it is a copy. If we updated go-ethereum-sonic to a newer go-ethereum version, we might easily miss if there are changes in this code.

I would prefer if we could add something to `go-ethereum-sonic` that makes this code re-usable. If, for instance, you need public access to the `testMatcher` and some of its functions, I would suggest the following:

- rename the file `tests/init_test.go` in `go-ethereum-sonic` to `tests/init_tests.go` or something else, just to get rid of the `test` suffix and to make the content part of the `tests` library
- create a file `tests/sonic_extension.go` in `go-ethereum-sonic`
- in this file, define a type `TestMatcher` that is an alias for `testMatcher` or a struct that just contains a `testMatcher` instance
- for this new type in `sonic_extensions.go` add all needed public methods calling the internal methods as needed

This should eliminate most of the need for duplicating the code. Wdyt?

tests/state_test.go
HerbertJordan, line 1: The same argument holds for this as for the `init_test.go`. It would be great if we could minimize the amount of code duplication between the repos to make upgrades easier.

HerbertJordan (changes requested):

tests/init_test.go
kjezek: this will work when we accept the method with `*testing.T` parameters are outside of the `_test.go` files. As discussed online, I had considered this as an anti pattern, but when we accept it, then yes, the solution is that straightforward.

tests/state_test.go
kjezek: in this case, I guess the simplest way forward is not to use the method `withTrace`, I do not know how it is useful, but I personally did not use this tracing while debuting the tests at all.  The rest of the implementation iterates all the tests and executes them, i.e. should be fine IMHO.

gossip/evmstore/carmen.go
kjezek: yes, logs are checked here for all tests: https://github.com/Fantom-foundation/go-ethereum-sonic/blob/eth1.14.8/tests/state_test_util.go#L229

(this is the reason I added this method)

HerbertJordan (approved):
Looks good, thanks!

PR #227, opened 2024-10-15
Empty parents for genesis events

Tjofil:
This PR makes all genesis events not have any parents. Having a parent by other creator as a genesis turns out to be problematic due to Lachesis' iterative frame calculation, thus inducing frame non-monotonicity for the event chains. Such non-monotonicity makes future incrementalization efforts impossible and potentially hides other performance and security concerns.

We also introduce a new flag used only for `events import` subcommand that instructs Lachesis to suppress the frame check panic. This may be useful for regression testing where the monotonicity fix from https://github.com/Fantom-foundation/lachesis-base-sonic/pull/3 is not important and could obstruct ingestion of historical data.


psubotic (approved):

psubotic:
document describing change: https://docs.google.com/document/d/1yIcoJooH4LQnsfd4Z670mfQPYl3R2XPOrbuNp6cq9Ug/edit?usp=sharing

psubotic (dismissed):

gossip/emitter/emitter_test.go
```diff
@@ -115,4 +122,17 @@ func TestEmitter(t *testing.T) {
...
+		if selfParent != nil {
+			t.Error("genesis event must not have self parent")
+		}
+		if len(parents) > 0 {
+			t.Error("genesis event must not have any parents")
+		}
+		if !ok {
+			t.Error("genesis parent assignment must succeed without any parents")
+		}
+	})
```
HerbertJordan, line 137: This is a rather unexpected location for a test that targets a function implemented in `parents.go`. By the usual convention, it should be placed in a file named `parents_test.go`.

The test also seems to be rather unrelated to any other test in this file, so factoring it out should be rather straight forward.

gossip/emitter/emitter_test.go
```diff
@@ -55,6 +56,12 @@ func TestEmitter(t *testing.T) {
 	external.EXPECT().StateDB().
 		Return(nil).
 		AnyTimes()
+	external.EXPECT().GetLastEvent(idx.Epoch(1), cfg.Validator.ID).
+		Return(nil).
+		Times(1)
+	external.EXPECT().GetHeads(idx.Epoch(1)).
+		Return(hash.Events{(&tdag.TestEvent{}).ID(), (&tdag.TestEvent{}).ID()}).
```
HerbertJordan, line 63: nit: no need to specify a return value if this function is supposed to never be called (`Times(0)`)

config/flags/flags.go
```diff
@@ -331,4 +331,10 @@ var (
 		Usage: "Password to unlock validator private key",
 		Value: "",
 	}
+	
+	// Consensus
+	SuppressFramePanicFlag = cli.BoolFlag{
+		Name:  "suppress-frame-panic",
```
HerbertJordan, line 337: It looks like the convention in this list of flags is to use a naming schema like `<component>.<property>` to group flags. The property is all-lower-case without spaces. According to this, this flag should be named something like `lachesis.suppressframepanic` or `lachesis.ignoreframemissmatch`. Due to the long name I would probably take `lachesis.ignore-frame-miss-match`.
Wdyt?

HerbertJordan (changes requested):
This PR covers two features:
 - the change making sure that every genesis event is free of any parent event link
 - the introduction of the `SuppressFramePanicFlag`

These changes are independent of each other and should be handled in two distinct PRs. Would you be fine splitting this PR up in two?

Tjofil:
> This PR covers two features:
> 
> * the change making sure that every genesis event is free of any parent event link
> * the introduction of the `SuppressFramePanicFlag`
> 
> These changes are independent of each other and should be handled in two distinct PRs. Would you be fine splitting this PR up in two?

We planned with the testing team to do it more quickly, in one go. The suppression flag is needed both because of this PR and https://github.com/Fantom-foundation/lachesis-base-sonic/pull/3, i.e. the suppression flag is a direct consequence of all the monotonicity fixing effort.

I can still make it two PRs if you prefer.

HerbertJordan:
> We planned with the testing team to do it more quickly, in one go. The suppression flag is needed both because of this PR and [Fantom-foundation/lachesis-base-sonic#3](https://github.com/Fantom-foundation/lachesis-base-sonic/pull/3), i.e. the suppression flag is a direct consequence of all the monotonicity fixing effort.
> 
> I can still make it two PRs if you prefer.

Typically, PRs should be focused on a single change. I think it would not hurt to split those up, and it should not take more than 10 minutes. I can review them quickly as they become available.

However, if keeping those two changes bundled up is fixing some integration test issue for you, its fine for me to keep it as it is.

PR #224, opened 2024-10-10
Delete Opera's EVM clone and replace it with parameterized go-ethereum EVM implementation

HerbertJordan:
This PR replaces the outdated customized copy of Ethereum's EVM implementation in Sonic with a parameterized version if the EVM introduced in go-ethereum-sonic ([#5](https://github.com/Fantom-foundation/go-ethereum-sonic/pull/5)).

In particular, the following elements are dropped and replaced by their go-ethereum counterpart:
- `evmcore.GasPool` => `core.GasPool`
- `evmcore.Message` => `*core.Message`
- `evmcore.ApplyMessage` => `core.ApplyMessage`

PR #231, opened 2024-10-16
Add an exception to the gas fee checks for internal transactions

HerbertJordan:
This PR disables `BaseFee` checks for internal transactions.

The check for base-fee limits was introduced by replacing the former Sonic EVM with a parameterized version of the go-ethereum EVM in #224. 

The former Sonic EVM was not supporting the gas-price model introduced by the London revision ([EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)). Thus, any gas fees suggested by internal transactions have been accepted. The new go-ethereum EVM, however, is enforcing those gas price limits unless explicitly disabled. Disabling this feature is for internal transactions is what this PR is adding.


thaarok (approved):
Great, thanks!

PR #238, opened 2024-10-17
Add txScrambler.

cabrador:
This PR adds func that sorts transactions order  using `XOR` on transaction hash. Furhtermore, if there are transactions from same sender, these transactions are sorted by the nonce.


gossip/tx_scrambler.go
```diff
@@ -0,0 +1,156 @@
...
+	hash   common.Hash
+	sender common.Address
+	nonce  uint64
+}
+
+type scramblerSortFunc func(entries []*scramblerEntry, seed []byte) []*scramblerEntry
+
+// getExecutionOrder first removes any entries with duplicate hashes, then sorts the list by XORed hashes.
+// Furthermore, if there are entries with same sender, these entries are sorted by their nonce (lower comes first).
+func getExecutionOrder(entries []*scramblerEntry, sortFunc scramblerSortFunc) []*scramblerEntry {
```
HerbertJordan, line 21: This implementation does not follow the specification in the doc. In particular, it is not deterministic. This test fails:
```
func TestGetExecutionOrder_SortIsDeterministic2(t *testing.T) {
	entryA := &scramblerEntry{
		hash: common.Hash{1},
	}
	entryB := &scramblerEntry{
		hash: common.Hash{2},
	}

	res1 := getExecutionOrder([]*scramblerEntry{entryA, entryB}, builtInSort)
	res2 := getExecutionOrder([]*scramblerEntry{entryB, entryA}, builtInSort)

	if !reflect.DeepEqual(res1, res2) {
		t.Fatal("slices have different order - algorithm is not deterministic")
	}
}
```

gossip/tx_scrambler.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+
+		// if we have not found all indexes of given address, append and continue
+		if len(idxs) != occurrence {
+			indexMap[entry.sender] = append(idxs, i)
+			continue
+		}
+	}
+
+	for _, idxs := range indexMap {
```
HerbertJordan, line 82: This is an `O(N^2)` algorithm -- we should try to limit it to `O(NlogN)`.

HerbertJordan (changes requested):

gossip/tx_scrambler_test.go
```diff
@@ -0,0 +1,215 @@
...
+	)
+	size := int64(10)
+
+	for _ = range numLoops {
+		b.Run(fmt.Sprintf("builtIn_%d", size), func(b *testing.B) {
+			for i := 1; i <= b.N; i++ {
+				b.StopTimer()
+				entries := createRandomScramblerTestInput()
+				b.StartTimer()
+				getExecutionOrder(entries)
```
HerbertJordan, line 130: The `size` is nowhere used in this benchmark ... it should be odd that the benchmark probably reports the same time for any size of input.

HerbertJordan (changes requested):
The algorithm looks good, but it should be extended to handle situations where the sender and nonce are equivalent, but the hash is different. Right now, we can make the tie-breaker the hash in this case, but in a follow-up we may want to use the gas price.

Besides that, I added a few comments on the testing and benchmarking. I think this needs another round of improvements.

gossip/tx_scrambler.go
```diff
@@ -0,0 +1,108 @@
...
+	slices.SortFunc(list, func(a, b *scramblerEntry) int {
+		aX = xorBytes32(a.hash, salt)
+		bX = xorBytes32(b.hash, salt)
+		return bytes.Compare(aX[:], bX[:])
+	})
+}
+
+// unifyEntries removes any transactions with duplicate hashes and creates the XOR salt from the unique tx list.
+// Furthermore, it returns whether given list of entries contains duplicate addresses.
+func unifyEntries(entries []*scramblerEntry) ([]*scramblerEntry, [32]byte, bool) {
```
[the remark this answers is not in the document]
cabrador: I didnt want to to split it since it can be done in one loop. I honestly like `analyseEntryList`

gossip/tx_scrambler_test.go
```diff
@@ -0,0 +1,215 @@
...
+
+	entries = getExecutionOrder(entries)
+
+	if got, want := len(entries), size/2; got != want {
+		t.Errorf("duplicate entries does not seem to be removed, got size: %v, want size: %v", got, want)
+	}
+}
+
+func TestTxScrambler_GetExecutionOrder_SortsSameSenderByNonce(t *testing.T) {
+	entries := createScramblerTestInputWithRepeatedAddresses()
```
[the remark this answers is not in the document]
cabrador: Added loop which increments number of repeated senders and the total number of entries.

gossip/tx_scrambler_test.go
```diff
@@ -0,0 +1,215 @@
...
+			if entries[i].sender == entries[j].sender {
+				if entries[i].nonce > entries[j].nonce {
+					t.Errorf("incorrect nonce order %d must be before %d", entries[j].nonce, entries[i].nonce)
+				}
+			}
+		}
+	}
+}
+
+func TestTxScrambler_GetExecutionOrder_SortIsDeterministic(t *testing.T) {
```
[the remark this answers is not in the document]
cabrador: Added two different tests. One for duplicate data and one for repeated data.
Both tests run in loop with increasing number of entries

gossip/tx_scrambler.go
```diff
@@ -0,0 +1,119 @@
...
+	// no need to sort more
+	if !hasDuplicateAddresses {
+		return uniqueList
+	}
+
+	return sortTransactionsByNonce(uniqueList)
+}
+
+// sortTransactionsByNonce finds any duplicate senders and sorts their transactions by nonce ascending.
+func sortTransactionsByNonce(entries []*scramblerEntry) []*scramblerEntry {
```
HerbertJordan, line 30: This operation is not deterministic. See for example this test case:
```

func TestTxScrambler_SortTransactionsByNonce_IsDeterministic(t *testing.T) {
	entries := []*scramblerEntry{}
	for i := range 10 {
		entries = append(entries, &scramblerEntry{
			hash: common.Hash{byte(i)},
		})
	}

	printOrder := func(entries []*scramblerEntry) string {
		builder := strings.Builder{}
		for _, entry := range entries {
			builder.WriteString(fmt.Sprintf("%d ", entry.hash[0]))
		}
		return builder.String()
	}

	want := sortTransactionsByNonce(entries)
	for range 10 {
		input := slices.Clone(want)
		shuffleEntries(input)
		got := sortTransactionsByNonce(input)
		if !reflect.DeepEqual(want, got) {
			t.Errorf("sort is not deterministic, want %v, got %v",
				printOrder(want),
				printOrder(got),
			)
		}
	}
}
```

HerbertJordan (changes requested):
The algorithm you have is not (yet) deterministic -- and although there are tests for it, they failed to detect this.

Please re-evaluate your testing strategy and try to make it ironclad.

gossip/tx_scrambler.go
HerbertJordan: Actually, forget that. I was writing this up too fast, I have to rethink this test case.

gossip/tx_scrambler_test.go
```diff
@@ -0,0 +1,342 @@
...
+		}
+		seenHashes[entry.hash] = true
+	}
+}
+
+func createRandomSalt() [32]byte {
+	var salt = [32]byte{}
+	for i := 0; i < 32; i++ {
+		salt[i] = byte(rand.Intn(math.MaxUint8))
+	}
```
HerbertJordan, line 340: Simpler:
```suggestion
	rand.Read(salt[:])
```

gossip/tx_scrambler_test.go
```diff
@@ -0,0 +1,342 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"golang.org/x/exp/rand"
+	"math"
+	"math/big"
+	"reflect"
+	"testing"
+)
+
+func TestTxScrambler_AnalyseEntryList_RemovesDuplicateTransactions(t *testing.T) {
+	entries := createScramblerTestInputRepeatedHashes(2, 10)
```
HerbertJordan, line 14: With these test input generation functions being out-factored and parameterized it is very difficult to understand what is actually tested in these functions. 

I would argue that in unit tests clarity is favorable over code re-use. It should be possible for anyone by just reading the code in the test to figure out what the test is about and whether the test itself is doing what the label promises.

A test like this would be much easier to comprehend:
```
func TestTxScrambler_AnalyseEntryList_RemovesDuplicateTransactions2(t *testing.T) {
	entries := []*scramblerEntry{
		{hash: common.Hash{2}},
		{hash: common.Hash{1}},
		{hash: common.Hash{1}},
	}
	entries, _, _ = analyseEntryList(entries)

	want := []*scramblerEntry{
		{hash: common.Hash{2}},
		{hash: common.Hash{1}},
	}
	if !reflect.DeepEqual(want, entries) {
		t.Errorf("want %v, got %v", want, entries)
	}
}
```

Or it might be even extended to multiple different input cases without harming readability a lot:
```
func TestTxScrambler_AnalyseEntryList_RemovesDuplicateTransactions3(t *testing.T) {
	examples := [][]*scramblerEntry{
		{},
		{
			{hash: common.Hash{1}},
		},
		{
			{hash: common.Hash{1}},
			{hash: common.Hash{1}},
		},
		{
			{hash: common.Hash{1}},
			{hash: common.Hash{1}},
			{hash: common.Hash{1}},
		},
		{
			{hash: common.Hash{1}},
			{hash: common.Hash{2}},
			{hash: common.Hash{1}},
		},
		{
			{hash: common.Hash{1}},
			{hash: common.Hash{2}},
			{hash: common.Hash{3}},
			{hash: common.Hash{2}},
			{hash: common.Hash{1}},
		},
	}

	for _, example := range examples {
		t.Run("", func(t *testing.T) {
			result, _, _ := analyseEntryList(example)

			seen := map[common.Hash]struct{}{}
			for _, entry := range result {
				if _, seen := seen[entry.hash]; seen {
					t.Fatalf("duplicate hash %v", entry.hash)
				}
				seen[entry.hash] = struct{}{}
			}
		})
	}
}
```

gossip/tx_scrambler_test.go
```diff
@@ -0,0 +1,342 @@
...
+	entries, _, _ = analyseEntryList(entries)
+
+	checkDuplicateHashes(t, entries)
```
HerbertJordan, line 17: nit: This test would pass if the implementation of `analyseEntryList` would just always return an empty list.

gossip/tx_scrambler_test.go
```diff
@@ -0,0 +1,342 @@
...
+	res1 := createScramblerTestInputRepeatedAddr(2, 10)
+	res2 := deepCopyEntries(res1)
+	// shuffle one array
+	shuffleEntries(res2)
+
+	for i := 0; i < 10; i++ {
+		salt := createRandomSalt()
+		scrambleTransactions(res1, salt)
+		for j := 0; j < 10; j++ {
+			scrambleTransactions(res2, salt)
```
HerbertJordan, line 57: I think `res2` should be shuffled each time before calling `scrambleTransactions`

gossip/tx_scrambler_test.go
```diff
@@ -0,0 +1,342 @@
...
+		b.Run(fmt.Sprintf("builtIn_%d", size), func(b *testing.B) {
+			for i := 1; i <= b.N; i++ {
+				b.StopTimer()
+				entries := createRandomScramblerTestInput(size)
+				b.StartTimer()
+				getExecutionOrder(entries)
+			}
+		})
+		size = size * multiplier
+	}
```
HerbertJordan, line 204: simpler:
```suggestion
	for size := uint64(10); size < 100_000; size *= 10 {
		b.Run(fmt.Sprintf("size_%d", size), func(b *testing.B) {
			for i := 1; i <= b.N; i++ {
				b.StopTimer()
				entries := createRandomScramblerTestInput(size)
				b.StartTimer()
				getExecutionOrder(entries)
			}
		})
	}
```

HerbertJordan (changes requested):
I think the algorithm is good, with the exception of how the same nonces are handled. This should be improved.

About the tests: it takes a lot of time to think them through due to the extra nesting of generator functions. The goal for tests is different than for regular code. In tests it should be plane obvious to the reader of the code what is going on, without knowing a lot of the other code around it. If several functions need to be studied in detail to understand a test, the test is too complicated.

The reasoning is that tests themself are not tested. Reviews are the last line of defense. You need to write them in a way that allows anybody to gain trust in the fact that these tests are doing what they are supposed to do.

Please give those tests another overhaul with the aim of making it easy for reviewers to understand what is going on.

gossip/tx_scrambler.go
```diff
@@ -0,0 +1,128 @@
...
+
+func xorBytes32(a, b [32]byte) (dst [32]byte) {
+	for i := 0; i < 32; i++ {
+		dst[i] = a[i] ^ b[i]
+	}
+	return
+}
+
+// deepCopyEntries returns deep copy of entries.
+func deepCopyEntries(entries []*scramblerEntry) []*scramblerEntry {
```
HerbertJordan, line 120: nit: only needed in tests => move to test file

gossip/tx_scrambler.go
```diff
@@ -0,0 +1,128 @@
...
+			return -1
+		}
+		// if nonce is equal, sort by gas
+		if a.gas > b.gas {
+			return -1
+		} else {
+			// if both nonce and gas is equal, order between those do not matter
+			return 1
+		}
+	})
```
HerbertJordan, line 52: To guarantee a total order, you need to consider the case where `a.sender == b.sender && a.nonce == b.nonce && a.gas == b.gas`; in this case, you should use the transaction hash as a tie breaker.

gossip/tx_scrambler.go
```diff
@@ -0,0 +1,128 @@
...
+			return 1
+		}
+		if a.nonce < b.nonce {
+			return -1
+		}
+		// if nonce is equal, sort by gas
+		if a.gas > b.gas {
+			return -1
+		} else {
+			// if both nonce and gas is equal, order between those do not matter
```
HerbertJordan, line 49: Attention: It does matter!

The sort function expects that `Cmp(a,b)==1` if and only if `Cmp(b,a)==-1`; If you have two equal elements `a` and `b` right now, you would return `1` in both cases.

gossip/tx_scrambler_test.go
```diff
@@ -0,0 +1,597 @@
...
+		{hash: common.Hash{3}},
+		{hash: common.Hash{2}},
+		{hash: common.Hash{1}},
+	}
+
+	shuffleEntries(entries)
+	result, _, _ := analyseEntryList(entries)
+	if len(result) == 0 {
+		t.Fatal("analyseEntryList returned empty list")
+	}
```
HerbertJordan, line 25: This can be stricter:
```suggestion
	if len(result) != 3 {
		t.Fatalf("unexpected length of result list, wanted 3, got %d", len(result))
	}
```

gossip/tx_scrambler_test.go
```diff
@@ -0,0 +1,597 @@
...
+			nonce:  2,
+		},
+		{
+			hash:   common.Hash{5},
+			sender: common.Address{1},
+			nonce:  2,
+		},
+	}
+
+	sortTransactionsWithSameSender(entries)
```
HerbertJordan, line 146: The input is already sorted in the desired way. So if this function does nothing, the test passes.

gossip/tx_scrambler_test.go
```diff
@@ -0,0 +1,597 @@
...
+	}
+
+	shuffleEntries(entries)
+	sortTransactionsWithSameSender(entries)
+	for i := 0; i < len(entries); i++ {
+		for j := i + 1; j < len(entries); j++ {
+			if entries[i].sender == entries[j].sender {
+				if entries[i].nonce > entries[j].nonce {
+					t.Errorf("incorrect nonce order %d must be before %d", entries[j].nonce, entries[i].nonce)
+				}
```
HerbertJordan, line 193: nit: seems to be irrelevant since all nonces are 1

gossip/tx_scrambler_test.go
```diff
@@ -0,0 +1,597 @@
...
+					hash:   common.Hash{3},
+					sender: common.Address{3},
+					nonce:  3,
+					gas:    3,
+				},
+				{
+					hash:   common.Hash{4},
+					sender: common.Address{2},
+					nonce:  2,
+					gas:    4,
```
HerbertJordan, line 441: If I understand the test-case setup correctly, this should be a 2 and in the entry below a 1

HerbertJordan (changes requested):
Looks good, but the gas-tie breaker still needs to be added.

gossip/tx_scrambler_test.go
cabrador: Added shuffle before calling `sortTransactionsWithSameSender`

gossip/tx_scrambler_test.go
cabrador: You are right, thanks

gossip/tx_scrambler.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"slices"
+)
+
+// scramblerEntry stores meta information about transaction for sorting and filtering them.
+type scramblerEntry struct {
+	hash   common.Hash    // transaction hash
+	sender common.Address // sender of the transaction
+	nonce  uint64         // transaction nonce
+	gas    uint64         // transaction gas
```
HerbertJordan, line 15: `gas` => `gasPrice`

gossip/tx_scrambler_test.go
```diff
@@ -0,0 +1,645 @@
...
+
+func TestTxScrambler_GetExecutionOrder_RandomInput(t *testing.T) {
+	// this tests these input sizes:
+	// 1, 4, 16, 64, 256, 1024
+	for i := int64(1); i <= 1024; i = i * 4 {
+		input := createRandomScramblerTestInput(i)
+		cpy := deepCopyEntries(input)
+		shuffleEntries(cpy)
+		input = filterAndOrderTransactions(input)
+		cpy = filterAndOrderTransactions(input)
```
HerbertJordan, line 575: Should this get `cpy` as an argument?
```suggestion
		cpy = filterAndOrderTransactions(cpy)
```

gossip/tx_scrambler_test.go
```diff
@@ -0,0 +1,342 @@
...
+			t.Fatal("found duplicate hash in entries")
+		}
+		seenHashes[entry.hash] = true
+	}
+}
+
+func createRandomSalt() [32]byte {
+	var salt = [32]byte{}
+	for i := 0; i < 32; i++ {
+		salt[i] = byte(rand.Intn(math.MaxUint8))
```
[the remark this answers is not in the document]
HerbertJordan: This problem still seems to exist.

gossip/tx_scrambler_test.go
HerbertJordan: This one as well ...

gossip/tx_scrambler_test.go
```diff
@@ -0,0 +1,645 @@
...
+	}
+	return salt
+}
+
+// deepCopyEntries returns deep copy of entries.
+func deepCopyEntries(entries []*scramblerEntry) []*scramblerEntry {
```
HerbertJordan, line 637: nit: I think nowhere in these tests you actually need a deep copy and a deep equality since all entries are all constant. Shallow copies of slices and comparisons would do the job.

gossip/tx_scrambler.go
```diff
@@ -0,0 +1,122 @@
...
+	return
+}
+
+func compareAsc[T cmp.Ordered](a T, b T) int {
+	return cmp.Compare(a, b)
+}
+
+func compareDesc[T cmp.Ordered](a T, b T) int {
+	return -cmp.Compare(a, b)
+}
```
HerbertJordan, line 122: nit: these functions only add minor benefits over just calling
```
cmp.Compare(a.nonce, b.nonce) // ascending order
```
or
```
cmp.Compare(b.nonce, a.nonce)  // descending order
```
where it is needed. Also, they are both only used once, so they there is no big utility in defining extra functions for this.

HerbertJordan (approved):
Looks good, only 1-2 minor issues and style comments. Then I think it is ready to merge.

gossip/tx_scrambler_test.go
cabrador: Sorry I thought I factored this in.

gossip/tx_scrambler_test.go
```diff
@@ -0,0 +1,680 @@
...
+
+// dummyScramblerEntry represents scramblery entry data used for testing
+type dummyScramblerEntry struct {
+	hash     common.Hash    // transaction hash
+	sender   common.Address // sender of the transaction
+	nonce    uint64         // transaction nonce
+	gasPrice uint64         // transaction gasPrice
+}
+
+func (s *dummyScramblerEntry) Unwrap() *types.Transaction {
```
HerbertJordan, line 662: What is this needed for?

HerbertJordan (approved):

gossip/tx_scrambler_test.go
cabrador: archaic code. Removed

HerbertJordan (approved):

Issue #7, opened 2023-06-08
Failed to start the JavaScript console: bignumber.js


hkalina:
When trying to attach opera console, it aways fails on revision 60cb9f71b26762be9abf7e0fad4297a19900d999:
```
$ build/opera attach http://localhost:18545/
Fatal: Failed to start the JavaScript console: bignumber.js: SyntaxError: bignumber.js: Line 5:1 Could not load source map: open doc/bignumber.js.map: no such file or directory
```

From my investigation this is caused by upgrade of
```
github.com/dop251/goja v0.0.0-20200721192441-a695b0cdd498 // indirect
```
to
```
github.com/dop251/goja v0.0.0-20220405120441-9037c2b61cbf // indirect
```
which have been [done in go-ethereum-substate](https://github.com/Fantom-foundation/go-ethereum-substate/commit/2053ea8166d43370f732fb6e754b7566decc6b2b?diff=unified#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R21). (When I do this upgrade on clean go-opera, it reproduce the same issue.)

However attempts to downgrade it are immediately reverted by `go mod tidy`

hkalina:
Resolved by removing the .map file reference from the .js file.

PR #244, opened 2024-10-22
Disable shutdown trigger when shutdown is complete

HerbertJordan:
This fix adds a termination to the go-routine handling shutdowns whenever pressing CTRL+C. Before this fix, the trigger kept running even after the shutdown was completed.

While running integration tests, it was noticed that the termination of one instance also caused a message about a repeated attempt of killing preceding instances emerged in the logs. Without this change, this would eventually put a limit of at most 10 integration tests to the infrastructure, since after that, the shutdown signal to one test would trigger an `exit` call to the first test that was run in the same process.

This is part of [#7](https://github.com/Fantom-foundation/sonic-admin/issues/7).


jenikd (approved):
Looks good thanks for that

PR #247, opened 2024-10-23
Cleanup error handling in ApplyTransaction unit test

HerbertJordan:
This PR adds error handling to a unit test where an error was quietly swallowed before.


LuisPH3 (approved):
I understand where does this come from. But from a quality point of view, a function returning err should not be trusted, including side effects to an external context. 
This kind of things may be unavoidable for this code base, but we should keep them in mind for Serse.

LuisPH3 (approved):

jenikd (approved):
Thanks

PR #251, opened 2024-10-23
Lint: Remove use of deprecated standard tools.

LuisPH3:
This PR removes the use of `ioutils`.
- `ioutil` replaced with a mix of `io` and `os`  symbols. [notes](https://stackoverflow.com/questions/75206234/for-go-ioutil-readall-ioutil-readfile-ioutil-readdir-deprecated)

### DIFF
--- a/gossip/filters/filter_test.go
+++ b/gossip/filters/filter_test.go
@@ -18,7 +18,6 @@ package filters
 
 import (
 	"context"
-	"io/ioutil"
 	"math/big"
 	"os"
 	"path"
@@ -55,7 +54,7 @@ func makeReceipt(addr common.Address) *types.Receipt {
 }
 
 func BenchmarkFilters(b *testing.B) {
-	dir, err := ioutil.TempDir("", "filtertest")
+	dir, err := os.MkdirTemp("", "filtertest")
 	if err != nil {
 		b.Fatal(err)
 	}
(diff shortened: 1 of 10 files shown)


gossip/filters/filter_test.go
```diff
@@ -55,7 +54,7 @@ func makeReceipt(addr common.Address) *types.Receipt {
 }
 
 func BenchmarkFilters(b *testing.B) {
-	dir, err := ioutil.TempDir("", "filtertest")
+	dir, err := os.MkdirTemp("", "filtertest")
```
HerbertJordan, line 57: How about using [b.TempDir()](https://pkg.go.dev/testing#B.TempDir) at this place? This would be a directory that is unique and automatically cleaned up at the end of the benchmark.

opera/genesisstore/fileshash/filehash_test.go
```diff
@@ -71,10 +70,10 @@ func TestFileHash_ReadWrite(t *testing.T) {
 
 func testFileHash_ReadWrite(t *testing.T, content []byte, expRoot hash.Hash, pieceSize uint64) {
 	require := require.New(t)
-	tmpDirPath, err := ioutil.TempDir("", "filehash*")
+	tmpDirPath, err := os.MkdirTemp("", "filehash*")
```
HerbertJordan, line 73: Also here, and in other places, [t.TempDir()](https://pkg.go.dev/testing#B.TempDir) would be the more canonical way to deal with temporary directories in tests.

HerbertJordan (changes requested):
Looks good, but I would suggest to use the canonical [TempDir()](https://pkg.go.dev/testing#T.TempDir) in tests whenever a temporary directory is required.

gossip/filters/filter_test.go
LuisPH3: You are right, for testing is even better to let the test tool handle temporaries

HerbertJordan (approved):
Thanks!

HerbertJordan (approved):

PR #252, opened 2024-10-23
Lint: Replace uses of bytes.Compare with bytes.Equal

LuisPH3:
This PR replaces uses of `Compare !=0` with `!bytes.Equal`.
Doing so is not only more idiomatic, but potentially faster. 


HerbertJordan (approved):
Thanks!

PR #253, opened 2024-10-23
Replace deprecated math/rand with pgregory.net/rand

LuisPH3:
This PR replaces `math/rand` with `pgregrory/rand`, Documentation [here](https://github.com/flyingmutant/rand)
This library was chosen, instead of `math/rand/v2` because of the extensive use of `rand.Read`. Rand read [has not been keep in V2](https://go.dev/blog/randv2) and a migration to `crypto/rand` would be needed.  Please notice that both libraries have incompatible API, each of them satisfying a subset of the functions existing in the deprecated `math/rand`.

`pgregrory/rand` has a compatible API except for the seeding mechanism, and migration was easier. This library has been used for quite some time in Tosca as well. 

### DIFF
--- a/cmd/sonictool/app/validator.go
+++ b/cmd/sonictool/app/validator.go
@@ -2,13 +2,14 @@ package app
 
 import (
 	"crypto/ecdsa"
-	"crypto/rand"
 	"errors"
 	"fmt"
 	"os"
 	"path"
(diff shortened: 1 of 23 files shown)


LuisPH3:
As discussed offline, This PR will be reworked.

PR #255, opened 2024-10-23
Add end-to-end BlobBaseFee instruction test

facuMH:
This PR adds an end-to-end test for BlobBaseFee opcode ([EIP-7516](https://eips.ethereum.org/EIPS/eip-7516))

- deploys a smart contract capable of retrieving the current blobbasefee for archive queries
- fetches the blobbasefee from the contract using a transaction (via a log message)
- fetches the blobbasefee from the archive using another method of the contract
- checks that the blobbasefee is zero and all sources report the same value. We compare to zero because BlobBaseFee is currently not supported by go-ethereum-sonic

This PR is part of [#10](https://github.com/Fantom-foundation/sonic-admin/issues/10)

### DIFF
--- a/tests/blobbasefee_test.go
+++ b/tests/blobbasefee_test.go
@@ -2,6 +2,7 @@ package tests
 
 import (
 	"context"
+	"math"
 	"testing"
 
 	"github.com/Fantom-foundation/go-opera/tests/contracts/blobbasefee"
@@ -72,9 +73,6 @@ func TestBlobBaseFee_CanReadBlobBaseFeeFromHeadAndBlockAndHistory(t *testing.T) 
 
 // helper functions to calculate blob base fee based on https://eips.ethereum.org/EIPS/eip-4844#gas-accounting
 func getBlobBaseFeeFrom(header *types.Header) uint64 {
-	// source for constants: https://eips.ethereum.org/EIPS/eip-4844#parameters
-	MIN_BASE_FEE_PER_BLOB_GAS := uint64(1)
-	BLOB_BASE_FEE_UPDATE_FRACTION := uint64(3338477)
 	blobGasUsed := uint64(0)
 	if header.BlobGasUsed != nil {
 		blobGasUsed = *header.BlobGasUsed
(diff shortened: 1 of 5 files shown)


tests/blobbasefee_test.go
HerbertJordan, line 1: Studying EIP-7516 it actually lists a method how to extract the blob-base-fee from a block header in its [Specification](https://eips.ethereum.org/EIPS/eip-7516#specification) section. It cites the formula for computing the blob-base-fee in [EIP-4844](https://eips.ethereum.org/EIPS/eip-4844#gas-accounting). This formula depends on [these](https://github.com/Fantom-foundation/go-ethereum-sonic/blob/2400937cc3b1c370ea8735d7af86eac621e4eb70/core/types/block.go#L89-L93) two fields in the block header.

Can you implement this formula in the test and verify that the header information is consistent with the blob basefee reported in transactions and archive queries?

HerbertJordan (changes requested):
Looks good, but I have an additional feature request.

HerbertJordan (dismissed):
Looks good, thanks!

jenikd:
Once PR https://github.com/Fantom-foundation/Sonic/pull/261 is merged, also comparison to the RPC function can be added, something like this:
```
	// call the blob base fee rpc method
	fromRpc := new(hexutil.Uint64)
	err = client.Client().Call(&fromRpc, "eth_blobBaseFee")
	if err != nil {
		t.Fatalf("failed to get blob base fee from rpc; %v", err)
	}
```

jenikd (dismissed):

jenikd (approved):
Thanks

tests/blobbasefee_test.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	// call the blob base fee rpc method
+	fromRpc := new(hexutil.Uint64)
+	err = client.Client().Call(&fromRpc, "eth_blobBaseFee")
+	if err != nil {
+		t.Fatalf("failed to get blob base fee from rpc; %v", err)
+	}
+
+	// we check blob base fee is zero because it is not implemented yet. TODO issue #147
```
HerbertJordan, line 68: This comment is out-dated.

HerbertJordan (approved):

HerbertJordan (approved):

PR #257, opened 2024-10-24
changes tmp dir creation to avoid too long names

kjezek:
This PR updates ethereum tests to create temp directory by `os.MkdirTemp` instead of standard `t.TempDir`. The reason is that ethereum tests produce extensively long names of tests. When `t.TempDir` is called on such a test, it fails on `file name too long` e.g.:

```
19:02:15                  state_test.go:71: TempDir: mkdir /tmp/TestStatePyspecscancuneip4844_blobspoint_evaluation_precompile_external_vectors.jsonsrcGeneralStateTestsFillerPyspecscancuneip4844_blobstest_point_evaluation_precompile.pytest_point_evaluation_precompile_external_vectorsfork_Cancun-state_test-versioned_hash_None-verify_kzg_proof_case_correct_proof_83e53423a2dd93feCancun03556998538: file name too long
```

### DIFF
--- a/tests/state_test.go
+++ b/tests/state_test.go
@@ -22,6 +22,7 @@ import (
 	"github.com/Fantom-foundation/Carmen/go/state/gostate"
 	"github.com/ethereum/go-ethereum/core/vm"
 	"github.com/ethereum/go-ethereum/tests"
+	"os"
 	"path/filepath"
 	"testing"
 )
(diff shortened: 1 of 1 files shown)


tests/state_test.go
```diff
@@ -84,6 +90,9 @@ func createCarmenFactory(t *testing.T) carmenFactory {
 		if err := st.Close(); err != nil {
 			t.Fatalf("cannot close state: %v", err)
 		}
+		if err := os.RemoveAll(dir); err != nil {
```
HerbertJordan, line 93: Can I ask you to move this into a `defer` right after the successful creation of the directory? This way, it is going to be removed even if the test fails.

HerbertJordan (approved):

tests/state_test.go
kjezek: ok, I moved it right after the directory was created. It must be in the `Cleanup` callback as the scope outlives this method.

HerbertJordan (approved):

PR #258, opened 2024-10-24
Add Transient Storage integration test

jenikd:
This PR adds integration test for Transient Storage opcode ([EIP-1153](https://eips.ethereum.org/EIPS/eip-1153))

deploys a smart contract capable of interacting with inner contract
inner contract implements transient variable which is set and obtained within same transaction thru main contract
main contract emits obtained value of transient variable of inner contract
checks, if the value stored during transaction and emited as log is not 0 
checks, if obtaining this value outside store function always returns 0
This PR is part of https://github.com/Fantom-foundation/sonic-admin/issues/22

### DIFF
--- a/tests/transientstorage_test.go
+++ b/tests/transientstorage_test.go
@@ -1,0 +1,68 @@
+package tests
+
+import (
+	"testing"
+
+	"github.com/Fantom-foundation/go-opera/tests/contracts/transientstorage"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+)
+
+func TestTransientStorage_TransientStorageIsValidInTransaction(t *testing.T) {
+	net, err := StartIntegrationTestNet(t.TempDir())
+	if err != nil {
+		t.Fatalf("Failed to start the fake network: %v", err)
+	}
+	defer net.Stop()
+
+	// Deploy the transient storage contract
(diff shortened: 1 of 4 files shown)


tests/transientstorage_test.go
```diff
@@ -0,0 +1,68 @@
...
+		t.Fatalf("failed to get transient value from archive at head time; %v", err)
+	}
+
+	// Check that all non log values are zero
+	if valueBefore.Sign() != 0 || fromArchive.Sign() != 0 || fromArchiveHead.Sign() != 0 {
+		t.Fatalf("unexpected value; expected 0, got valueBefore %v, fromArchive %v, FromArchiveHead %v", valueBefore, fromArchive, fromArchiveHead)
+	}
+
+	// Check that the log value is non-zero
+	if fromLog.Sign() == 0 {
```
HerbertJordan, line 65: nit: could be checked for having the value 42

HerbertJordan (approved):
Looks good, thanks!

tests/transientstorage_test.go
jenikd: Done

LuisPH3 (approved):

HerbertJordan (approved):

PR #261, opened 2024-10-24
Add eth_blobBaseFee RPC function

jenikd:
This PR adds an RPC function eth_blobBaseFee

For now, it only returns 1 as blobs are not implemented in Sonic client
As blobs are not supported yet, blob base fee is equal to min blob gas price, because calculation of blob base fee is based on the blob gas price and excess blob gas which is always 0

### DIFF
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -155,7 +155,10 @@ func (s *PublicEthereumAPI) EffectiveBaseFee(ctx context.Context) *hexutil.Big {
 }
 
 func (s *PublicEthereumAPI) BlobBaseFee(ctx context.Context) *hexutil.Big {
-	return (*hexutil.Big)(big.NewInt(0))
+	// As blobs are not supported yet, blob base fee is equal to min blob gas price
+	// because calculation of blob base fee is based on the blob gas price and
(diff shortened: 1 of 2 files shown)


ethapi/api.go
```diff
@@ -154,6 +154,13 @@ func (s *PublicEthereumAPI) EffectiveBaseFee(ctx context.Context) *hexutil.Big {
 	return (*hexutil.Big)(s.b.EffectiveMinGasPrice(ctx))
 }
 
+func (s *PublicEthereumAPI) BlobBaseFee(ctx context.Context) *hexutil.Big {
```
HerbertJordan, line 157: Is there a way to query this as part of the Blog-Base-Fee integration test? If so, this might be a task for a follow-up PR, linking these changes up. If we ever implement blobs, we would be reminded to change this function as well.

HerbertJordan (approved):

ethapi/api.go
jenikd: Actually I added comment to integration test PR, how to do it 
https://github.com/Fantom-foundation/Sonic/pull/255#issuecomment-2437482247

PR #263, opened 2024-10-25
Lint, Prevent use of nil variable

LuisPH3:
The following code may access the variable `e` member functions when being `nil`

```
	e := store.GetEventPayload(id)
	if e == nil {
		log.Crit("Block event not found", "event", id.String())
	}
	fullEvents[i] = e
	gasPowerUsedSum += e.GasPowerUsed()    // <= nil access
```

part of https://github.com/Fantom-foundation/sonic-admin/issues/25


HerbertJordan (approved):
Looks good, thanks.

Fyi: [log.Crit](https://github.com/Fantom-foundation/go-ethereum-sonic/blob/2400937cc3b1c370ea8735d7af86eac621e4eb70/log/root.go#L106) terminates the process.

PR #264, opened 2024-10-25
Lint: Prevent copy of mutexes

LuisPH3:
When passing an object by value, content member variables are copied into a new memory address. This includes the content of a mutex lock, effectively cloning the mutex into a new instance. 

This PR introduces fixes for two scenarios detected, where mutexes are being copied. 

Part of https://github.com/Fantom-foundation/sonic-admin/issues/25

### DIFF
--- a/utils/dbutil/threads/pool_test.go
+++ b/utils/dbutil/threads/pool_test.go
@@ -16,9 +16,11 @@ func TestMain(m *testing.M) {
 
 func TestThreadPool(t *testing.T) {
 
-	for name, pool := range map[string]ThreadPool{
-		"global": GlobalPool,
-		"local":  ThreadPool{},
+	pool := ThreadPool{}
+
+	for name, pool := range map[string]*ThreadPool{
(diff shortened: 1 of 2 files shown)


utils/dbutil/threads/pool_test.go
```diff
@@ -16,9 +16,11 @@ func TestMain(m *testing.M) {
...
 func TestThreadPool(t *testing.T) {
 
-	for name, pool := range map[string]ThreadPool{
-		"global": GlobalPool,
-		"local":  ThreadPool{},
+	pool := ThreadPool{}
+
+	for name, pool := range map[string]*ThreadPool{
+		"global": &GlobalPool,
+		"local":  &pool,
```
HerbertJordan, line 23: fyi: it is a bit strange, but this also works
```suggestion
		"local":  {},
```

thaarok (approved):

HerbertJordan (approved):

PR #266, opened 2024-10-29
Add PREVRANDAO calculation

cabrador:
This PR adds computation of `PREVRANDAO`.


inter/block.go
```diff
@@ -15,12 +15,31 @@ type Block struct {
...
+func (b *Block) GetPrevRandao() common.Hash {
+	if b.prevRandao == (common.Hash{}) {
+		b.computePrevRandao()
+	}
+	return b.prevRandao
+}
+
+// computePrevRandao computes the PrevRandao from transaction hashes.
+func (b *Block) computePrevRandao() {
+	for _, event := range b.Events {
```
HerbertJordan, line 35: You should reset `b.prevRandao` to zero before computing the xor of the events (conservative coding).

inter/block.go
```diff
@@ -15,12 +15,31 @@ type Block struct {
...
+		for i := 0; i < 24; i++ {
+			// first 8 bytes should be ignored as they are not pseudo-random.
+			b.prevRandao[i+8] = b.prevRandao[i+8] ^ event[i+8]
+		}
+	}
+}
```
HerbertJordan, line 41: In the proposal the xor-value over all event-suffices is hashed to expand the 24-byte value into a 32-byte value.

inter/block_test.go
```diff
@@ -0,0 +1,42 @@
...
+		{byte(rand.Int())},
+		{byte(rand.Int())},
+	}
+	blk := Block{Events: events}
+	randao1 := blk.GetPrevRandao()
+	rand.Shuffle(len(blk.Events), func(i, j int) {
+		blk.Events[i], blk.Events[j] = blk.Events[j], blk.Events[i]
+	})
+	randao2 := blk.GetPrevRandao()
+	if !reflect.DeepEqual(randao1, randao2) {
```
HerbertJordan, line 23:
```suggestion
	if randao1 != randao2 {
```

inter/block_test.go
```diff
@@ -0,0 +1,42 @@
...
+		t.Error("computation is not deterministic")
+	}
+}
+
+func TestBlock_ComputePrevRandao_First8BytesAreZero(t *testing.T) {
```
HerbertJordan, line 28: This is not intended, that's why the final hash is added in the proposal.

evmcore/evm.go
```diff
@@ -51,12 +51,10 @@ func NewEVMBlockContext(header *EvmHeader, chain DummyChain, author *common.Addr
 		baseFee = new(big.Int).Set(header.BaseFee)
 	}
 
-	// Before Sonic upgrade random is always empty
-	// and difficulty is not 0
 	difficulty := big.NewInt(1)
 	if header.PrevRandao.Cmp(common.Hash{}) != 0 {
 		random = &header.PrevRandao
-		difficulty.SetUint64(0)
+		difficulty.Set(random.Big())
```
HerbertJordan, line 57: If I read [EIP-4399](https://eips.ethereum.org/EIPS/eip-4399#using-mixhash-field-instead-of-difficulty) correctly, difficulty is supposed to be 0 after the Merge update (Paris).

evmcore/evm_test.go
HerbertJordan, line 1: Maybe this file should be deleted instead.

gossip/c_block_callbacks.go
```diff
@@ -183,7 +183,7 @@ func consensusCallbackBeginBlockFn(
 
 				// Execute pre-internal transactions
 				preInternalTxs := blockProc.PreTxTransactor.PopInternalTxs(blockCtx, bs, es, sealing, statedb)
-				preInternalReceipts := evmProcessor.Execute(preInternalTxs)
+				preInternalReceipts := evmProcessor.Execute(preInternalTxs, common.Hash{})
```
HerbertJordan, line 186: Instead of adding the prev-randao to the `Execute` function I think it would be a better fit to pass it to the `Start` function (for instance in line 181) instead, since all the other context information is also set there.

gossip/blockproc/evmmodule/evm.go
```diff
@@ -121,10 +119,7 @@ func (p *OperaEVMProcessor) Execute(txs types.Transactions) types.Receipts {
 }
 
 func (p *OperaEVMProcessor) Finalize() (evmBlock *evmcore.EvmBlock, skippedTxs []uint32, receipts types.Receipts) {
-	evmBlock = p.evmBlockWith(
-		// Filter skipped transactions. Receipts are filtered already
-		inter.FilterSkippedTxs(p.incomingTxs, p.skippedTxs),
-	)
+	evmBlock = p.evmBlockWith(inter.FilterSkippedTxs(p.incomingTxs, p.skippedTxs))
```
HerbertJordan, line 122: nit: why is this comment removed? Is it wrong?

gossip/c_block_callbacks.go
```diff
@@ -238,7 +238,10 @@ func consensusCallbackBeginBlockFn(
 						txs = append(txs, e.Txs()...)
 					}
 
-					_ = evmProcessor.Execute(txs)
+					prevRandao := block.GetPrevRandao()
+
+					// todo pass salt from scrambler if txcount > 1 to avoid unnecessary compua
```
HerbertJordan, line 243: This TODO is probably out-dated.

HerbertJordan (changes requested):
In generally, this looks good, but I would suggest the following design change: pass the prevrandao the EVM processor through its `Start` function, not the `Execute` function; this would avoid changing the `EVMProcessor` interface and all other block-related context details are also passed in through the `Start` function.

inter/block.go
cabrador: I dont think its necessary because this func gets called only if `prevrandao` is zero.
```// GetPrevRandao returns the PrevRandao for current block.
func (b *Block) GetPrevRandao() common.Hash {
	if b.prevRandao == (common.Hash{}) {
		b.computePrevRandao()
	}
	return b.prevRandao
}

But yeah, in future we might call this func somewhere else so its a good practice.

File: evmcore/evm.go (hunk 5 above)
Comment (cabrador) [reply evmcore/evm.go:57]: Thats right, although this set actually changes the `prevrandao` in `header` (the contract test passes with my change) and if I add this condition to the contract test, it actually passes 
```	
if block.Difficulty().Uint64() != 0 {
	t.Errorf("incorrect block difficulty: %d", block.Difficulty().Uint64())
}
```

Seems like some weird stuff happens behind the curtains...

File: gossip/c_block_callbacks.go (hunk 6 above)
Comment (cabrador) [reply gossip/c_block_callbacks.go:186]: That was my initial thought, but the necessary events are not yet ready. I might be missing something, but I dont think passing `prevrandao` to `Start` is possible...

File: gossip/blockproc/evmmodule/evm.go (hunk 7 above)
Comment (cabrador) [reply gossip/blockproc/evmmodule/evm.go:122]: Removed by mistake. Thanks

File: gossip/c_block_callbacks.go (hunk 6 above)
Comment (cabrador) [reply gossip/c_block_callbacks.go:186]: I believe lines 226-239 need to be executed before Im able to call `GetPrevRandao()` https://github.com/Fantom-foundation/Sonic/blob/8fb94a51d66590e10fac2c2100a4dbd784d0fa48/gossip/c_block_callbacks.go#L226

File: evmcore/evm.go (hunk 5 above)
Comment (cabrador) [reply evmcore/evm.go:57]: I believe this happens because `tosca` version in sonic develop is outdated.
Tosca has fixed this in [this](https://github.com/Fantom-foundation/Tosca/commit/d6d14c1b9515b7ae8cdda15b781405888ce84b6c) commit last week.
While we use [this](https://github.com/Fantom-foundation/Tosca/commit/aa99babe0a10a3facbaba4f67382379e568d44b3) version 3 weeks old.

Comment (cabrador) [reply evmcore/evm.go]: https://github.com/Fantom-foundation/Sonic/pull/273 solves this issue

File: evmcore/evm.go [hunk 9]
```diff
@@ -51,8 +51,6 @@ func NewEVMBlockContext(header *EvmHeader, chain DummyChain, author *common.Addr
 		baseFee = new(big.Int).Set(header.BaseFee)
 	}
 
-	// Before Sonic upgrade random is always empty
-	// and difficulty is not 0
```
thaarok, line 55: We can reformulate this to:
```
// For legacy Opera network, the difficulty is always 1 and random nil
```

gossip/blockproc/evmmodule/evm.go
```diff
@@ -69,10 +70,6 @@ func (p *OperaEVMProcessor) evmBlockWith(txs types.Transactions) *evmcore.EvmBlo
 		baseFee = nil
 	}
 
-	prevRandao := common.Hash{}
-	if p.net.Upgrades.Sonic {
-		prevRandao.SetBytes([]byte{1}) // TODO provide pseudorandom data?
-	}
```
thaarok, line 75: If you drop this condition, you will break the Opera chain compatibility. We will not be able to synchonize Opera mainnet or testnet anymore, which breaks significant part of our CI.
I recommend:
```
	prevRandao := common.Hash{}
	if p.net.Upgrades.Sonic {
		prevRandao = p.prevRandao
	}
```
Alternatively you need to add similar condition elsewhere.

thaarok (commented):
If you remove the Sonic flag condition, the opera chain compatibility will broken, which will break our release tests.

gossip/blockproc/evmmodule/evm.go
cabrador: Thanks. I will bring it back.

evmcore/evm.go
cabrador: Forgot to rebase against `develop` - it already contains the necessary fix.

gossip/c_block_callbacks.go
cabrador: Nevertheless, I would like to make a Norma run to doublecheck that the block processing works correctly

inter/block_test.go
```diff
@@ -0,0 +1,25 @@
...
+	blk := Block{Events: events}
+	randao1 := blk.GetPrevRandao()
+	rand.Shuffle(len(blk.Events), func(i, j int) {
+		blk.Events[i], blk.Events[j] = blk.Events[j], blk.Events[i]
+	})
+	randao2 := blk.GetPrevRandao()
+	if randao1 != randao2 {
+		t.Error("computation is not deterministic")
+	}
+}
```
HerbertJordan, line 25: Please also add a test case where for a given set of inputs (empty events, single event, 3 events) the correct hash is computed. The correct hash should be obtained in the test by re-performing the necessary operations.

tests/prevrandao_test.go
```diff
@@ -0,0 +1,71 @@
...
+		t.Fatalf("unexpected number of logs; expected 1, got %d", len(receipt.Logs))
+	}
+
+	entry, err := contract.ParseCurrentPrevRandao(*receipt.Logs[0])
+	if err != nil {
+		t.Fatalf("failed to parse log; %v", err)
+	}
+	fromLog := entry.Prevrandao
+
+	// Collect the prevrandao from the block header.
```
HerbertJordan, line 39: This comment is wrong. You are actually requesting the prevrandao from the archive. If you would request it from the block header, you would expect it to be different.

tests/prevrandao_test.go
```diff
@@ -0,0 +1,71 @@
...
+	client, err := net.GetClient()
+	if err != nil {
+		t.Fatalf("failed to get client; %v", err)
+	}
+	defer client.Close()
+
+	block, err := client.BlockByNumber(context.Background(), receipt.BlockNumber)
```
HerbertJordan, line 46: It would be great to add another section to this test that queries the random values from block 0 to the current head and check that the random values are all unique.

gossip/c_block_callbacks.go
```diff
@@ -222,22 +245,10 @@ func consensusCallbackBeginBlockFn(
 					// sort events by Lamport time
 					sort.Sort(confirmedEvents)
```
HerbertJordan, line 246: I think this sorting needs to be moved up together with the `&inter.Block{ ... }` initialization to have a canonical order of the confirmed events.

gossip/c_block_callbacks.go
```diff
@@ -222,22 +245,10 @@ func consensusCallbackBeginBlockFn(
...
 
-					// new block
-					var block = &inter.Block{
-						Time:    blockCtx.Time,
-						Atropos: cBlock.Atropos,
-						Events:  hash.Events(confirmedEvents),
-					}
 					for _, tx := range append(preInternalTxs, internalTxs...) {
```
HerbertJordan, line 248: I think with this change internal transactions get added to the end of the block instead of being placed at the head - right?

HerbertJordan (changes requested):

tests/prevrandao_test.go
cabrador: block 1 and 2 does not have any events hence it produces same hash - is this expected behaviour?

gossip/c_block_callbacks.go
cabrador: This change changes how they are orderer in the block, but the internal transactions are still executed first. I thought only the execution of the transactions matter - fixed.

cabrador: Transactions from events are actually never appended to the same array as internal transactions.
Internal transactions are added [here](https://github.com/Fantom-foundation/Sonic/blob/729e04882b9e3efd16e9a3f2b36807c40ffacbf0/gossip/c_block_callbacks.go#L232). Event transactions are just executed. 
So I guess my assumption that only order of `Execute(...)` calls matter should be correct? Anyhow, I still rearranged the code to append the internal transactions first.

gossip/c_block_callbacks.go
```diff
@@ -178,11 +178,52 @@ func consensusCallbackBeginBlockFn(
...
 				}
 
-				evmProcessor := blockProc.EVMModule.Start(blockCtx, statedb, evmStateReader, onNewLogAll, es.Rules, es.Rules.EvmChainConfig(store.GetUpgradeHeights()))
+				// --------------------------------------------------------------- //
+				// ------------------------- Prepare the block ------------------- //
+				// --------------------------------------------------------------- //
+				// 1) Pop both pre-internal and internal transactions
+				// 2) Add them to the block - they must be placed at the beginning
+				// 3) Sort events by lamport time
+				// 4) Exclude first events which exceed MaxBlockGas
```
HerbertJordan, line 187:
```suggestion
				// 4) Restrict list of events to no exceed MaxBlockGas
```

gossip/c_block_callbacks.go
```diff
@@ -178,11 +178,52 @@ func consensusCallbackBeginBlockFn(
...
+				// 5) Append event transactions
```
HerbertJordan, line 188:
```suggestion
				// 5) Assemble list of transactions in execution order
```

tests/prevrandao_test.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	if fromLog.Cmp(fromLatestBlock) != 0 {
+		t.Errorf("prevrandao mismatch; from log %v, from block %v", fromLog, fromLatestBlock)
+	}
+	if fromLog.Cmp(fromArchive) != 0 {
+		t.Errorf("prevrandao mismatch; from log %v, from archive %v", fromLog, fromArchive)
+	}
+
+	fromBlockThree, err := contract.GetPrevRandao(&bind.CallOpts{BlockNumber: big.NewInt(3)})
```
HerbertJordan, line 65: nit: It is a bit unmotivated why block 3 is picked here. Can you use `receipt.BlockNumber-1` here?

inter/block_test.go
```diff
@@ -0,0 +1,68 @@
...
+			events: []hash.Event{
+				{byte(rand.Int())},
+			},
+		},
+		{
+			name: "multiple_events",
+			events: []hash.Event{
+				{byte(rand.Int())},
+				{byte(rand.Int())},
+				{byte(rand.Int())},
```
HerbertJordan, line 50: Please use Event IDs with more than a single byte. Position 0 filled here is anyway ignored in the computation, since only positions 8-31 are used in the PrevRandao computaiton.

HerbertJordan (approved):
I think the c_block_callback is a good improvement. We should test it in Norma, and then it is good to be merged.

cabrador:
> I think the c_block_callback is a good improvement. We should test it in Norma, and then it is good to be merged.

Running [here](https://scala.fantom.network/job/Norma/job/Experimental/job/RunScenarioSonicParam/4/)

HerbertJordan (commented):
Looks good, waiting for the Norma test results with the approval.

jenikd:
When testing this branch when syncing mainnet, I observed an issue:
- it was able to sync **17 epochs** without any issue
- then in last block (96641301) of the epoch 315288 there is a difference
- second internal transaction has different **nonce**, **input data** and therefor **hash**
- because of that also **stateRoot** of this block is different
[block.json](https://github.com/user-attachments/files/17652657/block.json)

mainnet block for comparison can be obtained by:
`curl -s -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0x5C2A115",true],"id":1}' https://rpc.ftm.tools`

gossip/c_block_callbacks.go
HerbertJordan, line 1: Nice simplification!

inter/block.go
```diff
@@ -15,12 +16,33 @@ type Block struct {
...
+func (b *Block) SetPrevRandao(prevrandao common.Hash) {
+	b.prevRandao = prevrandao
+}
+
+func (b *Block) GetPrevRandao() common.Hash {
+	return b.prevRandao
+}
+
+// ComputePrevRandao computes the prevRandao from event hashes.
+func ComputePrevRandao(events []hash.Event) common.Hash {
```
HerbertJordan, line 35: nit: should this function be in the gossip package in its own file, since it is only used by `c_block_callbacks.go`?

inter/block.go
```diff
@@ -15,12 +16,33 @@ type Block struct {
...
+	prevRandao := common.Hash{}
+	for _, event := range events {
+		for i := 0; i < 24; i++ {
+			// first 8 bytes should be ignored as they are not pseudo-random.
+			prevRandao[i+8] = prevRandao[i+8] ^ event[i+8]
+		}
+	}
+	return sha256.Sum256(prevRandao.Bytes())
```
HerbertJordan, line 43: By the specification written up in docs, that also serve as the documentation for this feature, only the last 24 bytes are hashed. Here you are hashing all 32 bytes, including 8 leading 0s. I think we should fix this, because it  does not make a lot of sense to include some constant values in the hash.

inter/block_test.go
```diff
@@ -0,0 +1,55 @@
...
+	rand.Shuffle(len(events), func(i, j int) {
+		events[i], events[j] = events[j], events[i]
+	})
+	randao2 := ComputePrevRandao(events)
+	if randao1 != randao2 {
+		t.Error("computation is not deterministic")
+	}
+}
+
+func TestComputePrevRandao_ComputationWorks(t *testing.T) {
```
HerbertJordan, line 24:
```suggestion
func TestComputePrevRandao_ComputationProducesCorrectValue(t *testing.T) {
```

inter/block_test.go
```diff
@@ -0,0 +1,55 @@
...
+	tests := []struct {
```
HerbertJordan, line 25: In such a case, where we really want to check the correctness of the hashing function end-to-end I might opt to hard-code inputs and outputs. These results are never expected to change, so the test never needs to be updated (maintenance costs = 0).

As it is right now, it is a 1:1 copy of the implementation code and errors could be easily correlated between the implementation and the test case. If it would be a different implementation, it would be better, but like this this test is to easily missing systematic issues.

For computing reference values I suggest to hard-code random input events, then do the computation of the resulting hashes by hand and use some web-tool for computing the sha256 that is to be expected.

HerbertJordan (changes requested):
Looks good, thanks for simplifying it. However, there a few minor issues that should still be addressed.

inter/block.go
cabrador: I thought about the place for this func. Yeah, `gossip` package is the right place because of the reason you mentioned.

gossip/prevrandao.go
```diff
@@ -0,0 +1,21 @@
...
+// computePrevRandao computes the prevRandao from event hashes.
+func computePrevRandao(events []hash.Event) common.Hash {
+	bts := [24]byte{}
+	for _, event := range events {
+		for i := 0; i < 24; i++ {
+			// first 8 bytes should be ignored as they are not pseudo-random.
+			bts[i] = bts[i] ^ event[i+8]
+		}
+	}
+	fmt.Println(common.BytesToHash(bts[:]).String())
```
HerbertJordan, line 19: nit: debug code to be removed

HerbertJordan (dismissed):
Thanks!

cabrador:
Latest version currently fails a [unit test](https://scala.fantom.network/job/Norma/job/Experimental/job/RunScenarioSonicParam/) in Norma. Investigating....

HerbertJordan (dismissed):

kjezek:
I could add debug messages here and there and figured out that transaction was not processed because of:

`contract creation code storage out of gas`

from the EVM: https://github.com/Fantom-foundation/Sonic/blob/c/prevrandao/evmcore/state_processor.go#L181

not sure how this PR could change requirements for gas amount

cabrador:
I have also tested that this PR does not break the ability to sync with mainnet and the test has passed.

kjezek (changes requested):
I can confirm that the recent fix works, but I think it would be much cleaner to initialise `prevrandao` of a `block` as part of construction of this struct. 

At the moment it uses a setter which is called only in block processing, in other cases it computes `prevrandao` lazy as part of call to getter.  Since the `evnets`, which is the input of computation, is know at the creation of this struct, it would make the code cleaner to compute `prevrandao`  eagerly.  No setter is needed then. 

Also this design would limit chance of having the recent bug, because it would not allow for creating a block which is initialised differently based on whetter the setter is called or not.

cabrador:
> I can confirm that the recent fix works, but I think it would be much cleaner to initialise `prevrandao` of a `block` as part of construction of this struct.
> 
> At the moment it uses a setter which is called only in block processing, in other cases it computes `prevrandao` lazy as part of call to getter. Since the `evnets`, which is the input of computation, is know at the creation of this struct, it would make the code cleaner to compute `prevrandao` eagerly. No setter is needed then.
> 
> Also this design would limit chance of having the recent bug, because it would not allow for creating a block which is initialised differently based on whetter the setter is called or not.

Unfortunately this is not possible. First iteration of this PR did exactly what you suggest, however it was not possible because: 
a) we need to pass the `prevrandao` to the processor which would require to creating the block struct much sooner
b) creating the block struct sooner was a bit too complicated as it required a big refactoring of `c_block_callback`
c) it actually did not work - we could not sync with mainnet correctly

It is a bit too difficulty to explain here but basically I moved a lot of code and ended up scratching it due to the reasons above, I can tell you more on a meeting if you wish.

gossip/c_block_callbacks.go
```diff
@@ -178,7 +178,17 @@ func consensusCallbackBeginBlockFn(
 					}
 				}
 
-				evmProcessor := blockProc.EVMModule.Start(blockCtx, statedb, evmStateReader, onNewLogAll, es.Rules, es.Rules.EvmChainConfig(store.GetUpgradeHeights()))
+				prevRandao := inter.ComputePrevRandao(confirmedEvents)
```
HerbertJordan, line 181: You could move the initialization of the `block` up here and read the prevRandao value from the block. This eliminated the need for the public `ComputePrevRandao`, the `SetPrevRandao` on the struct, and would define a canonical source for this information.
```suggestion
				var block = &inter.Block{
					Time:    blockCtx.Time,
					Atropos: cBlock.Atropos,
					Events:  hash.Events(confirmedEvents),
				}
				prevRandao := block.GetPrevRandao()
```

inter/block.go
```diff
@@ -15,12 +15,25 @@ type Block struct {
...
+func (b *Block) SetPrevRandao(prevrandao common.Hash) {
+	b.prevRandao = prevrandao
+}
+
+func (b *Block) GetPrevRandao() common.Hash {
+	// Dummy blocks never run through block processing
+	if b.prevRandao == (common.Hash{}) {
+		b.prevRandao = ComputePrevRandao(b.Events)
+	}
+	return b.prevRandao
```
HerbertJordan, line 34: I would remove the `prevRandao` cache value and just always compute it based on `b.Events` when needed until I see this to be an actual cause for performance issues. The reason: the `Block` type is not protected against arbitrary modifications. Somebody may change the `Events` list without resetting `prevRandao` and it would be very difficult to figure out that these two fields are no longer aligned.

HerbertJordan (changes requested):
Thanks for identifying the issue causing Norma to fail.

inter/block.go
kjezek: I think that computing `prevrandao` as part of struct creation would work as well. The events are set when creating the struct, and I think they are not changed during the lifecycle of `block` instance.

kjezek:
> > I can confirm that the recent fix works, but I think it would be much cleaner to initialise `prevrandao` of a `block` as part of construction of this struct.
> > At the moment it uses a setter which is called only in block processing, in other cases it computes `prevrandao` lazy as part of call to getter. Since the `evnets`, which is the input of computation, is know at the creation of this struct, it would make the code cleaner to compute `prevrandao` eagerly. No setter is needed then.
> > Also this design would limit chance of having the recent bug, because it would not allow for creating a block which is initialised differently based on whetter the setter is called or not.
> 
> Unfortunately this is not possible. First iteration of this PR did exactly what you suggest, however it was not possible because: a) we need to pass the `prevrandao` to the processor which would require to creating the block struct much sooner b) creating the block struct sooner was a bit too complicated as it required a big refactoring of `c_block_callback` c) it actually did not work - we could not sync with mainnet correctly
> 
> It is a bit too difficulty to explain here but basically I moved a lot of code and ended up scratching it due to the reasons above, I can tell you more on a meeting if you wish.

I really cannot see the reason why this lazy cannot be remoed. In the code, it sets prevrando as a next line after struct creation, where ecactly the same events used for prevrando computation are used. 

image

In other cases, where the block is created, it gets a set of events and they are not changed during the lifecyckle of this object - if I got the code right. Prove me where this is not true, and than I would agree that an eager initialisation is a problem. Since your code computes `prevrandao` only lazyly once, even if you shown me that the events are updated, it would not recompute `prevrando`, i.e. would be the same as initialising it once at object creation.

inter/block.go
cabrador: Caching removed.

cabrador:
@kjezek @HerbertJordan You are right. For some reason from earlier versions of this PR I thought this was very hard to accomplish. Thanks for noticing.

inter/block.go
cabrador: As @HerbertJordan says, since the events are public, it can be changed hence the `Prevrandao` value would not correspond with it.

evmcore/dummy_block.go
```diff
@@ -49,7 +49,7 @@ type (
 
 		BaseFee *big.Int
 
-		PrevRandao common.Hash
+		PrevRandao common.Hash // == mixHash/mixDigest
```
kjezek, line 52: I am clueless what this added commnet say

kjezek (approved):
It looks good to me. 

It turns out that configuration of prevrandao determines which hardfork is enabled. I think it would be good to have a test for this as well.

evmcore/dummy_block.go
cabrador: prevRandao is often saved as a `mixHash` or `mixDigest`. If you have a look at the integration test, you can see I get the random value using `MixDigest()` method 
```
fromLatestBlock := block.MixDigest().Big()
```

HerbertJordan (approved):

PR #269, opened 2024-10-29
Add end-to-end test for invalid contract start character 0xEF

facuMH:
This PR adds an end-to-end test for invalid contracts whose code starts with `0xEF` ([EIP-3541](https://eips.ethereum.org/EIPS/eip-3541))

The test added tests this from two different mechanisms:
- deploys a smart contract that provides 4 functions which attempt to create an account with valid/invalid code (where invalid is code that starts with byte `0xEF`) using the instructions `CREATE` or `CREATE2`
- creates two custom transactions with the provided valid/invalid codes and attempts to run them.

For all cases with invalid codes, the receipt is checked for `types.ReceiptStatusFailed`.
For all cases  with valid codes, the receipt is checked for `types.ReceiptStatusSuccessful`.

This PR also makes use of `require` library for error and results verification. (previously used in https://github.com/Fantom-foundation/Sonic/blob/develop/config/config_custom_test.go#L51)

This PR is part of https://github.com/Fantom-foundation/sonic-admin/issues/19


HerbertJordan (dismissed):
Looks good, thanks!

tests/contracts/invalidstart/invalidstart.sol
```diff
@@ -0,0 +1,42 @@
...
+
+    function create2ContractWithInvalidCode() public {
+        create2OrRevert(invalidBytecode);
+    }
+
+    function create2ContractWithValidCode() public {
+        create2OrRevert(validBytecode);
+    }
+
+    function create2OrRevert(bytes memory bytecode) public {
```
LuisPH3, line 33: nit: private

tests/contracts/invalidstart/invalidstart.sol
```diff
@@ -0,0 +1,42 @@
...
+    bytes public validBytecode = hex"60fe60005360016000f3";
+    function createContractWithInvalidCode() public {
+        createOrRevert(invalidBytecode);
+    }
+
+    function createContractWithValidCode() public {
+        createOrRevert(validBytecode);
+    }
+
+    function createOrRevert(bytes memory bytecode) public {
```
LuisPH3, line 16: nit: private

tests/invalid_start_test.go
```diff
@@ -0,0 +1,94 @@
...
+	r.NoError(err, "failed to get nonce:")
+
+	price, err := client.SuggestGasPrice(context.Background())
+	r.NoError(err, "failed to get gas price:")
+	// ---------
+
+	transaction, err := types.SignTx(types.NewTx(&types.AccessListTx{
+		ChainID:  chainId,
+		Gas:      500_000, // some gas that is big enough to run the code
+		GasPrice: price,
```
LuisPH3, line 85: Not sure if we need gasPrice, I think that when it is  null it gets deduced.

tests/invalid_start_test.go
```diff
@@ -0,0 +1,94 @@
...
+		To:       nil,
+		Value:    big.NewInt(0),
```
LuisPH3, line 87: no need for value in this TX

LuisPH3 (approved):

tests/invalid_start_test.go
facuMH: without a price a `transaction underpriced` error occurs.

HerbertJordan: I do not think there is an auto-deduction of the price in this context. The `nil` value is likely treated like a zero.

HerbertJordan (approved):

LuisPH3 (changes requested):
Please refactor to use one single network

LuisPH3 (approved):

PR #274, opened 2024-10-30
Add end-to-end test for max init code size limit and meter

facuMH:
This PR adds an end-to-end test that stress the charging and limits of init code size ([EIP-3860](https://eips.ethereum.org/EIPS/eip-3860))

- deploys a smart contract capable attempts to create a new contract with the provided init code size, and reports how much this costed via a log message. 

- creates 3 contracts that initialize to 1, 2 and 3 32 byte words, and checks that the difference in costs is as expected.

- attempts to create a contract where the available gas is not enough to cover for the init code size.

- creates a contract using the max size for init code.

- attempts to create a contract using a code size larger than allowed with enough gas to pay for it. 

This PR is part of https://github.com/Fantom-foundation/sonic-admin/issues/21


tests/initcoder_test.go
```diff
@@ -0,0 +1,78 @@
...
+	r.NoError(err)
+	return log.Cost.Uint64(), receipt.GasUsed
+}
+
+func createContractWithCodeLenAndGas(r *require.Assertions, net *IntegrationTestNet, contract *initcode.Initcode, codeLen, gasLimit uint64) *types.Receipt {
+	receipt, err := net.Apply(func(opts *bind.TransactOpts) (*types.Transaction, error) {
+		opts.GasLimit = gasLimit
+		return contract.CreateContractWith(opts, big.NewInt(int64(codeLen)))
+	})
+	fmt.Printf("gas provided: %v, and gas used: %v\n", gasLimit, receipt.GasUsed) // THIS IS A DEBUG LINE
```
HerbertJordan, line 75: nit: remove

tests/initcoder_test.go
```diff
@@ -0,0 +1,78 @@
...
+	// Deploy the invalid start contract.
+	contract, receipt, err := DeployContract(net, initcode.DeployInitcode)
+	r.NoError(err)
+	r.Equal(types.ReceiptStatusSuccessful, receipt.Status, "failed to deploy contract")
+
+	const gasForOneWord uint64 = 54977
+
+	t.Run("create transaction with enough gas for init code", func(t *testing.T) {
+		// we need to provide same gas to all transactions so that we can compare the cost.
+		cost1Word, _ := createAndCheckContractWithCodeLenAndGas(r, net, contract, 30, gasForOneWord+4)
```
HerbertJordan, line 30: nit: can you include the case for 0 words? This would be an interesting edge case as well.

HerbertJordan, line 30: for documentation, it would be good to define a constant
```
const gasPerInitWord = 2
```
and use this constant here and below to document where the constants 2 and 4 are coming from.

tests/initcoder_test.go
```diff
@@ -0,0 +1,78 @@
...
+	})
+
+	t.Run("create transaction without enough gas for init code", func(t *testing.T) {
+		receipt = createContractWithCodeLenAndGas(r, net, contract, 42, gasForOneWord)
+		r.Equal(types.ReceiptStatusFailed, receipt.Status,
+			"unexpectedly succeeded to create contract without enough gas")
+	})
+
+	const MAX_INITCODE_SIZE = 49152
+	const MAX_INITCODE_COST = (MAX_INITCODE_SIZE / 32) * 2
```
HerbertJordan, line 48: also here, 2 should be replaced by a constant

tests/initcoder_test.go
```diff
@@ -0,0 +1,78 @@
...
+	const someCost = gasForOneWord + MAX_INITCODE_COST + 9_199
```
HerbertJordan, line 49: If I understand correctly, this should not be `someCost` but `sufficientGas`

HerbertJordan (changes requested):
The test seems to miss two additional ways of creating contracts:
 - creation using `create2` (likely to use the same code path as `create`, but in a black-box test like this one should not assume that)
 - creation of a contract using a transaction without a target; I assume also this should be limited and metered.

tests/contracts/initcode/initcode.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.8;
+
+contract InitCode {
+    event LogCost(uint256 cost);
+    
+    function createContractWith(uint codeSize) public {
+        uint256 before = gasleft();
+        assembly ("memory-safe") {
+            pop(create(0, 0, codeSize))
```
LuisPH3, line 10: This code is too obscure, can you document what is going on?

tests/initcoder_test.go
```diff
@@ -0,0 +1,78 @@
...
+	net, err := StartIntegrationTestNet(t.TempDir())
+	r.NoError(err)
+	defer net.Stop()
+
+	// Deploy the invalid start contract.
+	contract, receipt, err := DeployContract(net, initcode.DeployInitcode)
+	r.NoError(err)
+	r.Equal(types.ReceiptStatusSuccessful, receipt.Status, "failed to deploy contract")
+
+	const gasForOneWord uint64 = 54977
```
LuisPH3, line 26: where does this come from?

tests/initcoder_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	t.Run("create transaction with enough gas for init code", func(t *testing.T) {
+		// we need to provide same gas to all transactions so that we can compare the cost.
+		cost1Word, _ := createAndCheckContractWithCodeLenAndGas(r, net, contract, 30, gasForOneWord+4)
+		cost2Words, _ := createAndCheckContractWithCodeLenAndGas(r, net, contract, 42, gasForOneWord+4)
+		cost3Words, _ := createAndCheckContractWithCodeLenAndGas(r, net, contract, 90, gasForOneWord+4)
```
LuisPH3, line 32: this test should be a table

tests/initcoder_test.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+	})
+
+	t.Run("create transaction without enough gas for init code", func(t *testing.T) {
+		receipt = createContractWithCodeLenAndGas(r, net, contract, 42, gasForOneWord)
+		r.Equal(types.ReceiptStatusFailed, receipt.Status,
+			"unexpectedly succeeded to create contract without enough gas")
+	})
+
+	const MAX_INITCODE_SIZE = 49152
```
LuisPH3, line 47: mixing capiltalized and not capitalized symbols. Previous constant used other naming convention. Please use camelCase.

tests/initcoder_test.go
LuisPH3: where does 9199 come from?

tests/initcoder_test.go
LuisPH3, line 1: This file introduces a pattern :
```
call_test(delcaration)
call_test(delcaration)
call_test(delcaration)

 func call_test(..)  // test logic
```

But we write tests like this:
```
table = [
declaration
declaration
declaration
]

for test in table{
     // test logic
}
```

tests/initcoder_test.go
```diff
@@ -0,0 +1,78 @@
...
+func createAndCheckContractWithCodeLenAndGas(r *require.Assertions, net *IntegrationTestNet, contract *initcode.Initcode, codeLen, gasLimit uint64) (uint64, uint64) {
+	receipt := createContractWithCodeLenAndGas(r, net, contract, codeLen, gasLimit)
+	r.Equal(types.ReceiptStatusSuccessful, receipt.Status, "failed to create contract with code length %d", codeLen)
+
+	log, err := contract.ParseLogCost(*receipt.Logs[0])
+	r.NoError(err)
+	return log.Cost.Uint64(), receipt.GasUsed
+}
+
+func createContractWithCodeLenAndGas(r *require.Assertions, net *IntegrationTestNet, contract *initcode.Initcode, codeLen, gasLimit uint64) *types.Receipt {
```
LuisPH3, line 70: nit: the variable should always be called require, not r.

tests/initcoder_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"testing"
+
+	"github.com/Fantom-foundation/go-opera/tests/contracts/initcode"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestInitCodeSizeLimitAndMetered(t *testing.T) {
+	r := require.New(t)
```
LuisPH3, line 15: nit: the variable should be called require, not r

LuisPH3 (changes requested):
I think this test is weak, it depends too much on the code generation, which could easily change the static costs of the hard-coded variables.
An approach where different executions compare to each other could factor out implementation costs, leaving the difference between executions be the gas measurement wanted.

tests/initcoder_test.go
facuMH: I don't think so, because I need to compare the result of each call, I'm not only check that the calls are successful, but also cross checking between results.

tests/initcoder_test.go
facuMH: the idea here was to differentiate between constants taken from EIP, and constants that I made for the test.

tests/initcoder_test.go
facuMH: I'm changing the test to better show where this value comes from.

tests/initcoder_test.go
facuMH: I don't think table test style can be applied here because, for `"create transaction with enough gas for init code"` I want to compare each result, not just check that the deploy has been successful, while for the other 3 cases, 2 are fail check, and 1 is a regular success check.

tests/initcoder_test.go
facuMH: this number was part of the initialization cost of the init code size. it was been rework so that the numbers and intents are clearer.

tests/contracts/initcode/initcode.sol
```diff
@@ -0,0 +1,41 @@
...
+        emit LogCost(before - gasleft());
+    }
+
+    function create2ContractWith(uint codeSize) public {
+        // save gas before attempting to create
+        uint256 before = gasleft();
+        uint256 result = 0;
+        assembly  {
+            // top level expresssions are not supposed to return a value, but create does.
+            // so we need to call pop to catch this value.
```
LuisPH3, line 25: nit: we need to assign the value

LuisPH3 (approved):

LuisPH3 (dismissed):
:+1:

tests/initcoder_test.go
```diff
@@ -0,0 +1,157 @@
...
+		opts.GasLimit = gasLimit
+		return variant(opts, big.NewInt(int64(codeLen))) //
+	})
+	require.NoError(err)
+	return receipt
+}
+
+type variant func(opts *bind.TransactOpts, codeSize *big.Int) (*types.Transaction, error)
+
+func runTransactionWithCodeSizeAndGas(r *require.Assertions, net *IntegrationTestNet, codeSize, gas uint64) (*types.Receipt, error) {
```
LuisPH3, line 130: nit:
r => require

tests/initcoder_test.go
```diff
@@ -0,0 +1,157 @@
...
+	// these values are needed for the transaction but are irrelevant for the test
+	client, err := net.GetClient()
+	r.NoError(err, "failed to connect to the network:")
```
LuisPH3, line 133: The connection to the client should be closed. 
Just like other uses inside the integration net API.

defer client.Close()

facuMH: I closed it manually in line 144, because the call to `net.Run` in line 156 internally makes another call to `GetClient` and this sometimes resulted in a timeout. closing the client before calling run, stopped the timeouts. But I have to remove the commented defer, this was an oversight, sorry.

tests/initcoder_test.go
```diff
@@ -0,0 +1,156 @@
...
+		receipt2, err := runTransactionWithCodeSizeAndGas(require, net, 2, sufficientGas)
+		require.NoError(err)
+		require.Equal(types.ReceiptStatusSuccessful, receipt2.Status, "failed on transfer to empty receiver with valid code")
+
+		difference := receipt2.GasUsed - receipt1.GasUsed
+		require.Equal(difference, zeroByteCost, "gas difference between 1 and 2 words should be 4, instead got %d",
+			difference)
+	})
+
+	t.Run("make a transaction without enough gas for init code size", func(t *testing.T) {
```
LuisPH3, line 67: I don't understand this one. 
```
	transaction, err := types.SignTx(types.NewTx(&types.AccessListTx{
		ChainID:  chainId,
		Gas:      gas,
		GasPrice: price,
		To:       nil,
		Nonce:    nonce,
		Data:     make([]byte, codeSize),
```

This creates a transaction without code. and allocates a buffer for the data (input for the contract to be executed). It makes sense that it fails because of not having enough gas, but I do not understand what does this have to do with CREATE/CREATE2 
Can you develop?

tests/initcoder_test.go
```diff
@@ -0,0 +1,156 @@
...
+	require.NoError(err)
+	require.Equal(types.ReceiptStatusSuccessful, receipt.Status, "failed to deploy contract")
+
+	// deploy measureGasAndAssign to get cost of deploying a contract without create.
+	assignCost, gasUsed := createAndCheckContractWithCodeLenAndGas(require, net, contract, contract.MeasureGasAndAssign, 0, 0_000)
+
+	// -- using CREATE instruction
+	const wordCostCreate uint64 = 2
+	var gasForCreate uint64 = 32000 + gasUsed + assignCost
+	t.Run("create", func(t *testing.T) {
```
LuisPH3, line 36: because of variable name  - package name clash, it would be better to create the require variable inside of the function.

tests/initcoder_test.go
```diff
@@ -0,0 +1,156 @@
...
+		opts.GasLimit = gasLimit
+		return variant(opts, big.NewInt(int64(codeLen))) //
+	})
+	require.NoError(err)
+	return receipt
+}
+
+type variant func(opts *bind.TransactOpts, codeSize *big.Int) (*types.Transaction, error)
+
+func runTransactionWithCodeSizeAndGas(require *require.Assertions, net *IntegrationTestNet, codeSize, gas uint64) (*types.Receipt, error) {
```
LuisPH3, line 130: pass t and create require inside for better error location reporting.

tests/initcoder_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+func createAndCheckContractWithCodeLenAndGas(require *require.Assertions, net *IntegrationTestNet, contract *initcode.Initcode, variant variant, codeLen, gasLimit uint64) (uint64, uint64) {
+	receipt := createContractWithCodeLenAndGas(require, net, variant, codeLen, gasLimit)
+	require.Equal(types.ReceiptStatusSuccessful, receipt.Status, "failed to create contract with code length %d", codeLen)
+	log, err := contract.ParseLogCost(*receipt.Logs[0])
+	require.NoError(err)
+	return log.Cost.Uint64(), receipt.GasUsed
+}
+
+func createContractWithCodeLenAndGas(require *require.Assertions, net *IntegrationTestNet, variant variant, codeLen, gasLimit uint64) *types.Receipt {
```
LuisPH3, line 119: pass t and create require inside for better error location reporting.

tests/initcoder_test.go
```diff
@@ -0,0 +1,156 @@
...
+	})
+
+	t.Run("create transaction with MAX_INITCODE_SIZE+1", func(t *testing.T) {
+		receipt = createContractWithCodeLenAndGas(require, net, variant, MAX_INIT_CODE_SIZE+1, sufficientGas)
+		require.Equal(types.ReceiptStatusFailed, receipt.Status,
+			"unexpectedly succeeded to create contract with init code length greater than MAX_INITCODE_SIZE")
+	})
+}
+
+func createAndCheckContractWithCodeLenAndGas(require *require.Assertions, net *IntegrationTestNet, contract *initcode.Initcode, variant variant, codeLen, gasLimit uint64) (uint64, uint64) {
```
LuisPH3, line 111: pass t and create require inside for better error location reporting.

LuisPH3 (changes requested):
Unfortunatelly I do not understand the last 2 cases, I think they are not related to CREATE, and we may be testing something else.
Please take a look on them because if there is a property from CREATE that we should be testing, I do not think we manage to stimulate it via the transactions created.

tests/initcoder_test.go
```diff
@@ -0,0 +1,156 @@
...
+	})
+
+	t.Run("create transaction without enough gas for init code", func(t *testing.T) {
+		receipt = createContractWithCodeLenAndGas(require, net, variant, 30, gasForContract-wordCost)
+		require.Equal(types.ReceiptStatusFailed, receipt.Status,
+			"unexpectedly succeeded to create contract without enough gas")
+	})
+
+	t.Run("create transaction with max init code size", func(t *testing.T) {
+		_, _ = createAndCheckContractWithCodeLenAndGas(require, net, contract, variant, MAX_INIT_CODE_SIZE, sufficientGas)
```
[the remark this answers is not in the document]
facuMH: both returns are numerical values that are unrelated for this tests case. in this case we just want to check that it can run with those parameters and no error occurred. Errors are checked inside the function.

tests/initcoder_test.go
facuMH: this case is, as you mention, unrealted to CREATE/CREATE2. it is only testing that in a transaction where the init code is less than `MAX_INIT_CODE_SIZE`, it is charged for the of it.

LuisPH3: I get it now, the mentioned "create transaction". thanks it makes more sense now.

tests/initcoder_test.go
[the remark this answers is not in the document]
LuisPH3: errcheck will only report non-catched errors, other ignored return values can be ignored.

tests/initcoder_test.go
LuisPH3: as discussed offline, it is clear what this does not.

tests/initcoder_test.go
```diff
@@ -0,0 +1,156 @@
...
+	// -- using CREATE2 instruction
+	// create2 has extra costs for hashing, this is explained in
+	// https://eips.ethereum.org/EIPS/eip-3860 and reflected in evm.codes calculator.
+	const wordCostCreate2 uint64 = wordCostCreate + 6
+	var gasForCreate2 uint64 = gasForCreate + 44
+	t.Run("create2", func(t *testing.T) {
+		testForVariant(t, net, contract, receipt, contract.Create2ContractWith, gasForCreate2, wordCostCreate2)
+	})
+
+	t.Run("make a transaction that charges for the init code size", func(t *testing.T) {
```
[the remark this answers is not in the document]
LuisPH3: i see it now, these transactions create, follow semantics of the create operations, but they do not execute any create opcode

tests/initcoder_test.go
```diff
@@ -0,0 +1,156 @@
...
+		cost1Word, _ := createAndCheckContractWithCodeLenAndGas(require, net, contract, variant, 30, gasFor3Words)
+		cost2Words, _ := createAndCheckContractWithCodeLenAndGas(require, net, contract, variant, 42, gasFor3Words)
+		cost3Words, _ := createAndCheckContractWithCodeLenAndGas(require, net, contract, variant, 90, gasFor3Words)
+		require.Equal(cost2Words-cost1Word, wordCost,
+			"cost difference between 1 and 2 words should be %v, instead got %d", wordCost, cost2Words-cost1Word)
+		require.Equal(cost3Words-cost1Word, gasCostFor2Words,
+			"cost difference between 1 and 3 words should be %v, instead got %d", gasCostFor2Words, cost3Words-cost1Word)
+	})
+
+	t.Run("create transaction without enough gas for init code", func(t *testing.T) {
```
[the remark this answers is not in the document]
facuMH: the property I want to check here is not whether an account is created or not, but rather that "transaction fails without  enough gas". I will rename to that if you think is appropiate.

tests/initcoder_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	t.Run("aborts with init code size larger than MAX_INITCODE_SIZE", func(t *testing.T) {
+		receipt := createContractWithCodeLenAndGas(t, net, variant, MAX_INIT_CODE_SIZE+1, sufficientGas)
+		require := require.New(t)
+		require.Equal(types.ReceiptStatusFailed, receipt.Status,
+			"unexpectedly succeeded to create contract with init code length greater than MAX_INITCODE_SIZE")
+	})
+}
+
+func createAndCheckContractWithCodeLenAndGas(t *testing.T, net *IntegrationTestNet,
```
LuisPH3, line 120: nit: this function has too many responsibilities, and it feels difficult to read:
 I presume that this function was the result of a refactor that put together repeated code, nevertheless it lacks of a clear goal and responsibility. It just does too many things. 
 
I learned some rule, some long time ago: something like:  You do not do communication with the user inside the computation function, but instead return the result and write all IO (or stdout for the time being) at the same layer of the program. 

For this reason it would be better to have some pattern:
```
contract, err := createContract()
checkErr(err)
cost, err := getCost(contract)
checkErr(err)
```
Because in the domain of the unit test, error checking is reporting errors to the user. 
There are some reasons why this is this way, I will try to source some literature.

LuisPH3 (dismissed):

tests/initcoder_test.go
```diff
@@ -0,0 +1,170 @@
...
+	var gasForCreate2 uint64 = gasForCreate + 44
+	t.Run("create2", func(t *testing.T) {
+		testForVariant(t, net, contract, contract.Create2ContractWith, gasForCreate2, wordCostCreate2)
+	})
+
+	t.Run("create transaction", func(t *testing.T) {
+		t.Run("charges depending on the init code size", func(t *testing.T) {
+			// transactions charge 4 gas for each zero byte in data.
+			const zeroByteCost uint64 = 4
+			// create a transaction with 1 word of code.
```
HerbertJordan, line 53: It looks like the comments and error messages in this section are mixing up words and bytes ... please align them.

tests/initcoder_test.go
```diff
@@ -0,0 +1,170 @@
...
+			require.Equal(difference, zeroByteCost,
+				"gas difference between 1 and 2 words should be 4, instead got", difference)
+		})
+
+		t.Run("aborts with init code size larger than MAX_INITCODE_SIZE", func(t *testing.T) {
+			// as specified in https://eips.ethereum.org/EIPS/eip-3860#rules,
+			// this is similar to transactions considered invalid for not meeting the intrinsic gas cost requirement.
+			_, err := runTransactionWithCodeSizeAndGas(t, net, MAX_INIT_CODE_SIZE+1, sufficientGas)
+			require.ErrorContains(err,
+				"intrinsic gas too low", "unexpectedly succeeded to create contract with init code length greater than MAX_INITCODE_SIZE")
```
HerbertJordan, line 75: If the error is "intrinsic gas too low" then the transaction was not stopped by the code-size limit but by insufficient gas. I do not think that both of these errors should be accepted here. Only the second.

Is `require.ErrorContains` a either or or a all-of-those condition?

tests/initcoder_test.go
```diff
@@ -0,0 +1,170 @@
...
+	})
+
+	t.Run("create transaction", func(t *testing.T) {
+		t.Run("charges depending on the init code size", func(t *testing.T) {
+			// transactions charge 4 gas for each zero byte in data.
+			const zeroByteCost uint64 = 4
+			// create a transaction with 1 word of code.
+			receipt1, err := runTransactionWithCodeSizeAndGas(t, net, 1, sufficientGas)
+			require.NoError(err)
+			require.Equal(types.ReceiptStatusSuccessful, receipt1.Status,
```
HerbertJordan, line 56: If I understand the `require` library correctly, you should re-create a new `require` instance in every sub-test because otherwise `require` would report issues to the wrong `t *testing.T`.

tests/initcoder_test.go
```diff
@@ -0,0 +1,170 @@
...
+	const wordCostCreate2 uint64 = wordCostCreate + 6
+	var gasForCreate2 uint64 = gasForCreate + 44
+	t.Run("create2", func(t *testing.T) {
+		testForVariant(t, net, contract, contract.Create2ContractWith, gasForCreate2, wordCostCreate2)
+	})
+
+	t.Run("create transaction", func(t *testing.T) {
+		t.Run("charges depending on the init code size", func(t *testing.T) {
+			// transactions charge 4 gas for each zero byte in data.
+			const zeroByteCost uint64 = 4
```
HerbertJordan, line 52: It would be great to also check that non-zero values are correctly charged.

HerbertJordan (changes requested):

tests/initcoder_test.go
```diff
@@ -0,0 +1,170 @@
...
+const sufficientGas = uint64(100_000)
+
+func TestInitCodeSizeLimitAndMetered(t *testing.T) {
+	require := require.New(t)
+
+	net, err := StartIntegrationTestNet(t.TempDir())
+	require.NoError(err)
+	defer net.Stop()
+
+	// Deploy the invalid start contract.
```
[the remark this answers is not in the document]
facuMH: leftover comment from previous implementations. will be deleted.

tests/contracts/initcode/initcode.sol
```diff
@@ -0,0 +1,42 @@
...
+        emit LogCost(before - gasleft());
+    }
+
+    function Create2ContractWith(uint codeSize) public {
+        // save gas before attempting to create
+        uint256 before = gasleft();
+        uint256 result = 0;
+        assembly {
+            // top level expresssions are not supposed to return a value, but create does.
+            // so we need to assign its return value to catch it.
```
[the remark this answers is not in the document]
facuMH: I will change this comment to explain in a simpler manner why we need to store the result of create.

tests/initcoder_test.go
facuMH: to make this test, it would probably be best to make a different solidity contract. This might come later.

tests/initcoder_test.go
facuMH: as the comment above says, the first 2 rules in the EIP say that intrinsic gas error is the expected one. Or at least that was my interpretation. 

` ErrorContains(theError error, contains string, msgAndArgs ...interface{})` is the signature, I think this answers your quesiton.

tests/initcoder_test.go
facuMH: I hope the new changes move towards what you comment here

tests/initcoder_test.go
```diff
@@ -0,0 +1,184 @@
...
+	require.Equal(types.ReceiptStatusSuccessful, receipt.Status, "failed to create contract with code length ", codeLen)
+	return receipt
+}
+
+func createContractWithCodeLenAndGas(t *testing.T, net *IntegrationTestNet, variant variant, codeLen, gasLimit uint64) *types.Receipt {
+	receipt, err := net.Apply(func(opts *bind.TransactOpts) (*types.Transaction, error) {
+		opts.GasLimit = gasLimit
+		return variant(opts, big.NewInt(int64(codeLen)))
+	})
+	if err != nil {
```
LuisPH3, line 141: nit: mixing require and testing apis

tests/initcoder_test.go
```diff
@@ -0,0 +1,184 @@
...
+		t.Errorf("failed to create contract: %v", err)
+	}
+	return receipt
+}
+
+func getCreateCost(t *testing.T, contract *contractcreator.Contractcreator, receipt *types.Receipt) uint64 {
+	log, err := contract.ParseLogCost(*receipt.Logs[0])
+	if err != nil {
```
LuisPH3, line 149: nit:  mixing require and testing apis

tests/initcoder_test.go
```diff
@@ -0,0 +1,184 @@
...
+	contract, receipt, err := DeployContract(net, contractcreator.DeployContractcreator)
+	requireBase.NoError(err)
+	requireBase.Equal(types.ReceiptStatusSuccessful, receipt.Status, "failed to deploy contract")
+
+	// deploy measureGasAndAssign to get cost of deploying a contract without create.
+	receipt = createContractSuccessfully(t, net, contract.GetOverheadCost, 0, 0)
+
+	// -- using CREATE instruction
+	const wordCostCreate uint64 = 2
+	var gasForCreate uint64 = 32035 + receipt.GasUsed //+ assignCost
```
LuisPH3, line 35: nit: this comment is not clear, is this already added, or is it a comment to the formula?

tests/initcoder_test.go
```diff
@@ -0,0 +1,184 @@
...
+	// create2 has extra costs for hashing, this is explained in
+	// https://eips.ethereum.org/EIPS/eip-3860 and reflected in evm.codes calculator.
+	const wordCostCreate2 uint64 = wordCostCreate + 6
+	var gasForCreate2 uint64 = gasForCreate
+	t.Run("create2", func(t *testing.T) {
+		testForVariant(t, net, contract, contract.Create2With, gasForCreate2, wordCostCreate2)
+	})
+
+	t.Run("create transaction", func(t *testing.T) {
+		t.Run("charges depending on the init code size", func(t *testing.T) {
```
LuisPH3, line 50: nit: this case could be a function, like the others.
That would solve the alias problem with requireBase and the nested require variables.

tests/initcoder_test.go
facuMH: did not think it was worth to instantiate a new `require` just for this one simple check.

jenikd (changes requested):

tests/initcoder_test.go
```diff
@@ -0,0 +1,191 @@
...
+	var gasForCreate uint64 = 32035 + receipt.GasUsed
+	t.Run("create", func(t *testing.T) {
+		testForVariant(t, net, contract, contract.CreatetWith, gasForCreate, wordCostCreate)
+	})
+
+	// -- using CREATE2 instruction
+	// create2 has extra costs for hashing, this is explained in
+	// https://eips.ethereum.org/EIPS/eip-3860 and reflected in evm.codes calculator.
+	const wordCostCreate2 uint64 = wordCostCreate + 6
+	var gasForCreate2 uint64 = gasForCreate
```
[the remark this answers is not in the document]
facuMH: yes, this is a remnant from some experimentation I was doing, should have been removed.

jenikd (dismissed):
Thanks

tests/initcoder_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+		// createa a transaction with 2 byte of code.
+		receipt2, err := runTransactionWithCodeSizeAndGas(t, net, 2, sufficientGas)
+		require.NoError(err)
+		require.Equal(types.ReceiptStatusSuccessful, receipt2.Status,
+			"failed on transfer to empty receiver with valid code")
+
+		difference := receipt2.GasUsed - receipt1.GasUsed
+		require.Equal(difference, zeroByteCost,
+			"gas difference between 1 and 2 words should be 4, instead got", difference)
```
HerbertJordan, line 126: In the rest of this test you state that code size is measured in bytes, but here it is words. I guess it should be bytes here as well.

tests/initcoder_test.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func createContractWithCodeLenAndGas(net *IntegrationTestNet, variant variant, codeLen, gasLimit uint64) (*types.Receipt, error) {
+	return net.Apply(func(opts *bind.TransactOpts) (*types.Transaction, error) {
+		opts.GasLimit = gasLimit
+		return variant(opts, big.NewInt(int64(codeLen)))
+	})
+}
+
+func getCreateCost(contract *contractcreator.Contractcreator, receipt *types.Receipt) (uint64, error) {
```
HerbertJordan, line 157: nit: this is only used once ... inline?

HerbertJordan (dismissed):
Looks good, thanks.

HerbertJordan (approved):

PR #275, opened 2024-10-31
Add RPC response size limit for debug and trace namespace

jenikd:
This PR adds functionality created in PR #193 which was not merged into develop branch yet.

Refers to issue [#56](https://github.com/Fantom-foundation/sonic-admin/issues/56)

### DIFF
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -1901,13 +1901,14 @@ func (s *PublicTransactionPoolAPI) Resend(ctx context.Context, sendArgs Transact
 // PublicDebugAPI is the collection of Ethereum APIs exposed over the public
 // debugging endpoint.
 type PublicDebugAPI struct {
-	b Backend
+	b               Backend
+	maxResponseSize int // in bytes
 }
 
 // NewPublicDebugAPI creates a new API definition for the public debug methods
 // of the Ethereum service.
-func NewPublicDebugAPI(b Backend) *PublicDebugAPI {
-	return &PublicDebugAPI{b: b}
+func NewPublicDebugAPI(b Backend, maxResponseSize int) *PublicDebugAPI {
+	return &PublicDebugAPI{b: b, maxResponseSize: maxResponseSize}
 }
 
 // GetBlockRlp retrieves the RLP encoded for of a single block.
@@ -2071,7 +2072,7 @@ func (api *PublicDebugAPI) traceTx(ctx context.Context, tx *types.Transaction, m
 		return nil, err
 	}
 
-	if responseSizeLimit > 0 && len(result) > responseSizeLimit {
+	if api.maxResponseSize > 0 && len(result) > api.maxResponseSize {
 		return nil, ErrMaxResponseSize
 	}
 
@@ -2145,7 +2146,7 @@ func (api *PublicDebugAPI) traceBlock(ctx context.Context, block *evmcore.EvmBlo
 
 		// limit the response size.
 		resultsLength += len(res)
-		if responseSizeLimit > 0 && resultsLength > responseSizeLimit {
+		if api.maxResponseSize > 0 && resultsLength > api.maxResponseSize {
 			return nil, ErrMaxResponseSize
 		}
 
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -18,6 +18,7 @@ package ethapi
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"math/big"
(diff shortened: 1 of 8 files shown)


ethapi/limit.go
```diff
@@ -0,0 +1,91 @@
...
+	return nil
+}
+
+// writeString appends the contents of s to the buffer and handle possible panic
+func (b *JsonResultBuffer) writeString(s string) (err error) {
+	defer func() {
+		if recover() != nil {
+			err = ErrResponseTooLarge
+		}
+	}()
```
HerbertJordan, line 88: It might be better to check the limit first (`b.Len() + len(s) > responseSizeLimit`) and abort. Catching a panic in case of an out-of-memory allocation in this case is likely not a great idea since if memory is already so tight, chances are that a panic may also occur in any other part of the code in any other go-routine.

The idea with catching the panic should in my opinion be abandoned in favor of a reasonable default upper limit.

ethapi/limit_test.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	obj := testStruct{Str: "test string"}
+
+	err = b.AddObject(obj)
+	if err != ErrMaxResponseSize {
+		t.Errorf("expected ErrMaxResponseSize, got %v", err)
+	}
+}
+
```
HerbertJordan, line 102: If the panic-handling is retained, and this remains a key feature of the `JsonResultBuffer`, it should be tested.

gossip/config.go
```diff
@@ -86,6 +86,9 @@ type (
 		// allows only for EIP155 transactions.
 		AllowUnprotectedTxs bool
 
+		// MaxResponseSize is a limit for maximum response size in some RPC calls
```
HerbertJordan, line 89:
```suggestion
		// MaxResponseSize is a limit for maximum response size in some RPC calls in bytes
```

HerbertJordan (changes requested):
I would like to ask you to eliminate global configuration parameters like the size of the response buffer in favor of local instance variables.

Also, I think the idea of recovering from a panic in case of an out-of-memory issue should be  rethought and replaced in favor of preventing out-of-memory issues in the first place by checking limits before adding new results.

ethapi/api.go
```diff
@@ -2064,7 +2065,17 @@ func (api *PublicDebugAPI) traceTx(ctx context.Context, tx *types.Transaction, m
...
 		return nil, fmt.Errorf("tracing failed: %w", err)
 	}
-	return tracer.GetResult()
+
+	result, err := tracer.GetResult()
+	if err != nil {
+		return nil, err
+	}
+
+	if responseSizeLimit > 0 && len(result) > responseSizeLimit {
```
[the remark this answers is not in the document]
jenikd: Yes, this is exactly like you described it. I know, that this is not ideal, but at least it doesn't spend time trying to send response over the network

ethapi/limit.go
```diff
@@ -0,0 +1,91 @@
...
+
+// JsonResultBuffer is a bytes buffer for jsonRawMessage result
+type JsonResultBuffer struct {
+	bytes.Buffer
+}
+
+// NewJsonResultBuffer creates new bytes buffer
+func NewJsonResultBuffer() (b *JsonResultBuffer, err error) {
+	defer func() {
+		if recover() != nil {
```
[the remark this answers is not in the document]
jenikd: Because it tries to increase the buffer size`b.Grow(bufferStartSize)` to `bufferStartSize` not to grow in small increases if the defaul buffer is small, to avoid allocations for growing.

gossip/config.go
jenikd: Done

ethapi/limit.go
jenikd: Added checking max size before trying to increase the buffer, so if the limit is set reasonably, then panic should not occur. I still want to catch it in case of wrong limit was set.

ethapi/limit.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	if responseSizeLimit > 0 && b.Len()+len(res) > responseSizeLimit {
+		return ErrMaxResponseSize
+	}
+	if b.Len() > 1 {
+		if err := b.writeString(","); err != nil {
+			return err
+		}
+	}
+	if err = b.writeString(string(res)); err != nil {
```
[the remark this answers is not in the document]
jenikd: Done

ethapi/limit.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+// SetResponseSizeLimit sets maximum size of RPC response in bytes
+func SetResponseSizeLimit(limit int) {
+	responseSizeLimit = limit
+}
+
+// JsonResultBuffer is a bytes buffer for jsonRawMessage result
+type JsonResultBuffer struct {
+	bytes.Buffer
```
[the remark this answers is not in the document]
jenikd: Fixed

ethapi/limit.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+const (
+	// bufferStartSize default buffer start size in bytes
+	bufferStartSize = 10 * 1024
+)
+
+var (
+	// responseSizeLimit limits maximum size of RPC response
+	responseSizeLimit int
```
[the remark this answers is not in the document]
jenikd: Done

ethapi/limit_test.go
jenikd: I removed recovers and added checks for max result size before trying to grow buffer

HerbertJordan (approved):

PR #278, opened 2024-11-01
Integrate tx scrambler

cabrador:
This PR integrates previously created `txScrambler` into block processing.

### DIFF
--- a/gossip/c_block_callbacks.go
+++ b/gossip/c_block_callbacks.go
@@ -2,6 +2,8 @@ package gossip
 
 import (
 	"fmt"
+	"github.com/Fantom-foundation/go-opera/utils/signers/gsignercache"
+	"math/big"
 	"sort"
 	"sync"
 	"sync/atomic"
@@ -179,14 +181,14 @@ func consensusCallbackBeginBlockFn(
 				}
 
 				prevRandao := computePrevRandao(confirmedEvents)
-
+				chainCfg := es.Rules.EvmChainConfig(store.GetUpgradeHeights())
 				evmProcessor := blockProc.EVMModule.Start(
 					blockCtx,
 					statedb,
 					evmStateReader,
 					onNewLogAll,
 					es.Rules,
-					es.Rules.EvmChainConfig(store.GetUpgradeHeights()),
+					chainCfg,
 					prevRandao,
 				)
 				executionStart := time.Now()
@@ -260,14 +262,17 @@ func consensusCallbackBeginBlockFn(
 					sort.Sort(confirmedEvents)
 
 					blockEvents := spillBlockEvents(store, confirmedEvents, maxBlockGas)
-					txs := make(types.Transactions, 0, blockEvents.Len()*10)
+					unorderedTxs := make(types.Transactions, 0, blockEvents.Len()*10)
 					for _, e := range blockEvents {
-						txs = append(txs, e.Txs()...)
+						unorderedTxs = append(unorderedTxs, e.Txs()...)
 					}
 
-					for i, receipt := range evmProcessor.Execute(txs) {
+					signer := gsignercache.Wrap(types.MakeSigner(chainCfg, new(big.Int).SetUint64(number), uint64(blockCtx.Time)))
+					orderedTxs := getExecutionOrder(unorderedTxs, signer, es.Rules.Upgrades.Sonic)
+
+					for i, receipt := range evmProcessor.Execute(orderedTxs) {
 						if receipt != nil { // < nil if skipped
-							blockBuilder.AddTransaction(txs[i], receipt)
+							blockBuilder.AddTransaction(orderedTxs[i], receipt)
 						}
 					}
 
--- a/gossip/tx_scrambler.go
+++ b/gossip/tx_scrambler.go
@@ -45,8 +45,8 @@ func (tx *scramblerTransaction) Sender() common.Address {
 
 // getExecutionOrder returns correct order of the transactions.
 // If Sonic is enabled, the tx scrambler is used, otherwise the
-// order stays unchanged. If signer is unable to create sender for
-// a transaction, this transaction is not added to the final list.
+// order stays unchanged. If signer is unable to derive sender for
+// a transaction, this transaction is not excluded from the final list.
 func getExecutionOrder(unorderedTxs types.Transactions, signer types.Signer, isSonic bool) types.Transactions {
 	// Don't use scrambler if Sonic is not enabled
 	if !isSonic {
--- a/gossip/tx_scrambler.go
+++ b/gossip/tx_scrambler.go
@@ -174,18 +174,3 @@ func xorBytes32(a, b [32]byte) (dst [32]byte) {
 	}
 	return
 }
-
-// getSkippedTxNumbersWithinEvents takes skipped tx numbers by EVM and returns its original tx number before scrambling.
-func getSkippedTxNumbersWithinEvents(skippedByEvm []uint32, originalOrder map[common.Hash]uint32, orderedTxs types.Transactions) []uint32 {
-	originalOrderSkippedTxs := make([]uint32, 0, len(orderedTxs))
-	// Block needs the tx number from event, not from scrambler
-	for _, txNumber := range skippedByEvm {
-		// Find transaction index from the scrambled list
-		txHash := orderedTxs[txNumber].Hash()
-		// Find the original transaction index
-		originalTxNumber := originalOrder[txHash]
-		originalOrderSkippedTxs = append(originalOrderSkippedTxs, originalTxNumber)
-	}
-
-	return originalOrderSkippedTxs
-}
--- a/gossip/tx_scrambler.go
+++ b/gossip/tx_scrambler.go
@@ -3,7 +3,6 @@ package gossip
 import (
 	"bytes"
 	"cmp"
-	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"math/big"
 	"slices"
--- a/gossip/tx_scrambler.go
+++ b/gossip/tx_scrambler.go
@@ -3,6 +3,7 @@ package gossip
 import (
 	"bytes"
 	"cmp"
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"math/big"
 	"slices"
@@ -13,7 +14,7 @@ type ScramblerEntry interface {
 	// Hash returns the transaction hash
 	Hash() common.Hash
 	// Sender returns the sender of the transaction
-	Sender() *common.Address
+	Sender() common.Address
 	// Nonce returns the transaction nonce
 	Nonce() uint64
 	// GasPrice returns the transaction gas price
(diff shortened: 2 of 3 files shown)


gossip/tx_scrambler.go
```diff
@@ -15,8 +19,46 @@ type ScramblerEntry interface {
...
+	txs := make(types.Transactions, len(sorted))
+	for idx, e := range sorted {
+		txs[idx] = e.Unwrap()
+	}
+
+	return txs
+}
+
+// newScramblerTransaction creates a wrapper around *types.Transaction which implements ScramblerEntry.
+func newScramblerTransaction(signer types.Signer, tx *types.Transaction) ScramblerEntry {
```
HerbertJordan, line 40: This function actually only requires the sender of the transaction and does not need to check the signature. So it would be enough to just pass the sender.

API functions should always try to minimize what they are asking for. In this case, this just means that the signature check moves to the call side, but in different cases the sender may by available through another mechanism.

gossip/tx_scrambler.go
```diff
@@ -15,8 +19,46 @@ type ScramblerEntry interface {
 	Sender() common.Address
 	// Nonce returns the transaction nonce
 	Nonce() uint64
-	// GasPrice returns the transaction gasPrice price
-	GasPrice() uint64
+	// GasPrice returns the transaction gas price
+	GasPrice() *big.Int
+	// Unwrap returns the wrapped transaction.
+	Unwrap() *types.Transaction
```
HerbertJordan, line 25: This is not a good interface design. In most contexts where you are using the `ScrambleEntry` interface right now you do not need the `Unwrap` function -- this is even more evident that one of the implementations of this function states "//unused".

Instead of just adding this method for convenience to an existing interface you should stay true to the intention of the interface. The `ScrambleEntry` is intended to provide an abstraction for everything that can be scrambled. If the function `FilterAndOrderTransactions` needs something more, create a new interface that extends this one.

gossip/tx_scrambler.go
```diff
@@ -15,8 +19,46 @@ type ScramblerEntry interface {
...
+}
+
+// FilterAndOrderTransactions filters out duplicate transactions and sorts them.
+func FilterAndOrderTransactions(entries []ScramblerEntry) types.Transactions {
```
HerbertJordan, line 29: I would try to unify this: this function should have the same input and output type. Either the input and output is a slice of `ScramblerEntry` or a list of transactions. Right now, half the Transaction -> ScramblerEntry conversion is done by the caller and the other half by this function.

HerbertJordan (changes requested):

gossip/tx_scrambler.go
cabrador: I could not find any other way to derive the sender from the transaction. I will search again....

cabrador: I added a cache wrapper for the `Signer` developed by @thaarok - I believe that should solve that issue, is that correct?

gossip/tx_scrambler.go
cabrador: I agree this was not a good design. I removed the `Unwrap()` func and casted the interface back to the transactions.
My thoughts were that this file needs quite an extensive refactoring anyway, so casting is the easiest we can do rn and IMO its quite well readable.

When we start to refactor this, we might want to consider wrap the `types.Transactions` into an interface which actually implements `ScramblerEntry` so that we avoid some unnecessary looping.
Wdyt?

gossip/tx_scrambler.go
cabrador: This func was removed.

gossip/tx_scrambler.go
HerbertJordan: You have just developed this file and it already needs extensive refactoring? Why not do it right now, before the integration?

cabrador: I meant that we might refactor `gossip/c_block_callbacks.go` even more, or do you reckon the changes in #266 should be enough?

gossip/tx_scrambler.go
cabrador: We had a conversation with @jenikd - there seems there is no other option to derive the sender from transaction. Using the cached signer is the correct method.

cabrador:
Norma run passed [here](https://scala.fantom.network/job/Norma/job/Experimental/job/RunScenarioSonicParam/40/)

gossip/c_block_callbacks.go
```diff
@@ -233,11 +237,25 @@ func consensusCallbackBeginBlockFn(
...
-					txs := make(types.Transactions, 0, blockEvents.Len()*10)
+					unorderedTxs := make([]ScramblerEntry, 0, blockEvents.Len()*10)
+
+					signer := gsignercache.Wrap(types.MakeSigner(chainCfg, blockNumber, uint64(blockCtx.Time)))
 					for _, e := range blockEvents {
-						txs = append(txs, e.Txs()...)
+						for _, tx := range e.Txs() {
+							entry, err := newScramblerTransaction(signer, tx)
+							if err != nil {
+								log.Crit(fmt.Sprintf("cannot create scrambler tx %s", tx.Hash()), "err", err)
```
HerbertJordan, line 247: Please handle this without crashing the system. `log.Crit` will terminate the process.

gossip/c_block_callbacks.go
```diff
@@ -233,11 +237,25 @@ func consensusCallbackBeginBlockFn(
...
+							unorderedTxs = append(unorderedTxs, entry)
+						}
 					}
 
+					orderedTxs := getExecutionOrder(unorderedTxs, es.Rules.Upgrades.Sonic)
+					txs := make(types.Transactions, len(orderedTxs))
+					for i, tx := range orderedTxs {
+						// Cast back the transactions to pass it to the processor
+						txs[i] = tx.(*scramblerTransaction).Transaction
+					}
```
HerbertJordan, line 258: Ideally, this block is outlined into an helper function in `tx_scrambler.go` with a signature like this:
```
func orderTransactions(transactions types.Transactions, rules opera.Rules) types.Transactions
```
This will hide the implementation details of the `ScramblerEntry`, which could help simplify code and makes this function more readable.

gossip/tx_scrambler_test.go
```diff
@@ -600,6 +601,51 @@ func TestTxScrambler_FilterAndOrderTransactions_RandomInput(t *testing.T) {
...
+			name:    "TxScramblerIsNotUsed",
+			isSonic: false,
+			entries: []ScramblerEntry{
+				&dummyScramblerEntry{
+					hash: common.Hash{1},
+				},
+				&dummyScramblerEntry{
+					hash: common.Hash{1},
+				},
+			},
```
HerbertJordan, line 621: nit: this input is the same for all cases; Thus, I would move it into the test loop below.

HerbertJordan (changes requested):
Please consider reducing the amount of code needed in `c_block_callbacks.go` by encapsulating most functionality into a well-defined helper function handling the transaction shuffling. See the comments.

gossip/c_block_callbacks.go
cabrador: Im not sure how? This func does not return an error. If an error occurs in this part of the app, its always a fatal one...

cabrador: Discussed in private. Understood now

gossip/c_block_callbacks.go
cabrador: I wanted to avoid adding another loop, but I agree it will look better.

gossip/c_block_callbacks.go
```diff
@@ -262,6 +268,14 @@ func consensusCallbackBeginBlockFn(
...
 						}
 					}
+					// Skipped txs require original order from events
+					skippedTxs := make([]uint32, 0, len(orderedTxs))
+					for _, txNumber := range skippedByEvm {
+						txHash := orderedTxs[txNumber].Hash()
+						skippedTxs = append(skippedTxs, txPositions[txHash].EventOffset)
+					}
+					// Txs skipped by scrambler holds on the original order from events
+					block.SkippedTxs = append(skippedTxs, skippedByScrambler...)
```
HerbertJordan, line 278: Any idea how to test this? I think this code is too critical and complex to go without testing ...

gossip/c_block_callbacks.go
```diff
@@ -262,6 +268,14 @@ func consensusCallbackBeginBlockFn(
 							}
 						}
 					}
+					// Skipped txs require original order from events
+					skippedTxs := make([]uint32, 0, len(orderedTxs))
```
HerbertJordan, line 272: nit: the upper bound for this list should be the unordered list, right?

gossip/c_block_callbacks.go
```diff
@@ -262,6 +268,14 @@ func consensusCallbackBeginBlockFn(
...
+					for _, txNumber := range skippedByEvm {
+						txHash := orderedTxs[txNumber].Hash()
+						skippedTxs = append(skippedTxs, txPositions[txHash].EventOffset)
```
HerbertJordan, line 275: The `EventOffset` is, I think, the position of the transaction in the list of transactions of the first event where it is showing up. For the resulting skip list should reference the position of skipped transactions in the list of transactions one would obtain by merging the internal transactions, followed by explicit transactions, followed by the transactions from the events. An example use is [here](https://github.com/Fantom-foundation/Sonic/blob/668e297b7a2aa24b8c538f0ebaf2c4cf4d1b1684/gossip/evmstore/store_tx.go#L55).

gossip/c_block_callbacks.go
cabrador: right. Thanks

gossip/c_block_callbacks.go
cabrador: I guess we can create a helper func `getSkippedTransactions` and test it through that.

gossip/c_block_callbacks.go
```diff
@@ -262,6 +268,14 @@ func consensusCallbackBeginBlockFn(
...
 						}
 					}
+					// Skipped txs require original order from events
+					skippedTxs := make([]uint32, 0, len(unorderedTxs))
+					for _, txNumber := range skippedByEvm {
+						txHash := orderedTxs[txNumber].Hash()
+						skippedTxs = append(skippedTxs, txPositions[txHash].EventOffset)
+					}
+					// Txs skipped by scrambler holds on the original order from events
+					block.SkippedTxs = append(skippedTxs, skippedByScrambler...)
```
thaarok, line 278: @HerbertJordan, do you think it is a good idea to add another mechanism for skipping txs? Would not be more straightforward to let scrambler only reorder txs, with no filtering, and let only the evm to skip duplicitous nonces?

cabrador: After discussion with @thaarok and @jmpike we agreed that skipping the transactionss at this point is not necessary as the EVM will handle it. It would be diffiicult to include transactions for which we cannot derive the address into the `skippedTransactions`.
So I will remove this code.

HerbertJordan: @thaarok that was the original idea. The scrambler should only reorder the transactions. But then it turned out that there could be errors during decoding some of those transactions, during what is now the scrambling. Those transactions should be skipped. So it evolved into this code.

To clean this up, I think the best approach is to get rid of the list of skipped transactions all-together. This could come for free while doing the block-structure change we discussed this morning. Will know more in 1-2 days.

gossip/tx_scrambler_test.go
```diff
@@ -601,6 +604,114 @@ func TestTxScrambler_FilterAndOrderTransactions_RandomInput(t *testing.T) {
...
+			// one transaction is removed from the list
+			if got, want := len(txs), test.expectedLen; got != want {
+				t.Errorf("unexpected number of transasctions, got: %d, want: %d", got, want)
+			}
+		})
+	}
+
+}
+
+type errorEmittingSigner struct {
```
HerbertJordan, line 655: nit: you could have gotten this automatically using gomock ....

HerbertJordan (approved):
Looks good, thanks!

HerbertJordan (approved):
Much better, thanks!

cabrador:
[Norma](https://scala.fantom.network/job/Norma/job/Experimental/job/RunScenarioSonicParam/50/) passed

thaarok (approved):

PR #282, opened 2024-11-05
Withdrawals verification test

facuMH:
This PR adds a test that aims to verify the `Withdrawals` field added by ([EIP-4895](https://eips.ethereum.org/EIPS/eip-4895)).

The test:
- sets up a network, looks at the first block and checks that both, the `Withdrawals` list and the Header's `WithdrawalsHash` are empty or nil.
- creates a new block with an empty list of `Withdrawals`, runs RLP enconde/decode and then checks that the decoded list exists but is nil, and the `WithdrawalsHash` is the same as hashing an empty string. 

NOTE: during the development of this test I realized that by default blocks have a `nil` `Withdrawals` slice, which means that the header's `WithdrawalsHash` will be `nil` as well, and this is generates an error when decoding the field, since it expects a `[32]byte` but instead finds an empty string. Should we make it so that the default value for `Withdrawals` is not `nil`, but rather an empty list ? 

This PR contributes to [#57](https://github.com/Fantom-foundation/sonic-admin/issues/57).

### DIFF
--- a/evmcore/dummy_block.go
+++ b/evmcore/dummy_block.go
@@ -45,6 +45,8 @@ type (
 		GasLimit uint64
 		GasUsed  uint64
 
+		WithdrawalsHash *common.Hash
+
 		BaseFee *big.Int
 
 		PrevRandao common.Hash
@@ -85,16 +87,22 @@ func ToEvmHeader(block *inter.Block, index idx.Block, prevHash hash.Event, rules
 		prevRandao.SetBytes([]byte{1}) // TODO provide pseudorandom data?
 	}
 
+	withdrawalsHash := common.Hash{}
+	if rules.Upgrades.Sonic {
+		withdrawalsHash = types.EmptyWithdrawalsHash
+	}
+
 	return &EvmHeader{
-		Hash:       common.Hash(block.Atropos),
-		ParentHash: common.Hash(prevHash),
-		Root:       common.Hash(block.Root),
-		Number:     big.NewInt(int64(index)),
-		Time:       block.Time,
-		GasLimit:   math.MaxUint64,
-		GasUsed:    block.GasUsed,
-		BaseFee:    baseFee,
-		PrevRandao: prevRandao,
+		Hash:            common.Hash(block.Atropos),
+		ParentHash:      common.Hash(prevHash),
+		Root:            common.Hash(block.Root),
+		Number:          big.NewInt(int64(index)),
+		Time:            block.Time,
+		GasLimit:        math.MaxUint64,
+		GasUsed:         block.GasUsed,
+		BaseFee:         baseFee,
+		PrevRandao:      prevRandao,
+		WithdrawalsHash: &withdrawalsHash,
 	}
 }
 
@@ -102,17 +110,18 @@ func ToEvmHeader(block *inter.Block, index idx.Block, prevHash hash.Event, rules
 func ConvertFromEthHeader(h *types.Header) *EvmHeader {
 	// NOTE: incomplete conversion
 	return &EvmHeader{
-		Number:     h.Number,
-		Coinbase:   h.Coinbase,
-		GasLimit:   math.MaxUint64,
-		GasUsed:    h.GasUsed,
-		Root:       h.Root,
-		TxHash:     h.TxHash,
-		ParentHash: h.ParentHash,
-		Time:       inter.FromUnix(int64(h.Time)),
-		Hash:       common.BytesToHash(h.Extra),
-		BaseFee:    h.BaseFee,
-		PrevRandao: h.MixDigest,
+		Number:          h.Number,
+		Coinbase:        h.Coinbase,
+		GasLimit:        math.MaxUint64,
+		GasUsed:         h.GasUsed,
+		Root:            h.Root,
+		TxHash:          h.TxHash,
+		ParentHash:      h.ParentHash,
+		Time:            inter.FromUnix(int64(h.Time)),
+		Hash:            common.BytesToHash(h.Extra),
+		BaseFee:         h.BaseFee,
+		PrevRandao:      h.MixDigest,
+		WithdrawalsHash: h.WithdrawalsHash,
 	}
 }
 
@@ -136,6 +145,8 @@ func (h *EvmHeader) EthHeader() *types.Header {
 
 		Difficulty: new(big.Int),
 		MixDigest:  h.PrevRandao,
+
+		WithdrawalsHash: h.WithdrawalsHash,
 	}
 	// ethHeader.SetExternalHash(h.Hash) < this seems to be an optimization in go-ethereum-substate; skipped for now, needs investigation
 	return ethHeader
@@ -193,7 +204,7 @@ func (h *EvmHeader) ToJson(receipts types.Receipts) *EvmHeaderJson {
 		TotalDiff:       new(hexutil.Big),
 		Hash:            &h.Hash,
 		Epoch:           hexutil.Uint64(hash.Event(h.Hash).Epoch()),
-		WithdrawalsHash: &types.EmptyWithdrawalsHash,
+		WithdrawalsHash: h.WithdrawalsHash,
 		BlobGasUsed:     (*hexutil.Uint64)(new(uint64)),
 		ExcessBlobGas:   (*hexutil.Uint64)(new(uint64)),
 	}
--- a/evmcore/dummy_block.go
+++ b/evmcore/dummy_block.go
@@ -87,9 +87,9 @@ func ToEvmHeader(block *inter.Block, index idx.Block, prevHash hash.Event, rules
 		prevRandao.SetBytes([]byte{1}) // TODO provide pseudorandom data?
 	}
 
-	withdrawalsHash := common.Hash{}
+	var withdrawalsHash *common.Hash = nil
 	if rules.Upgrades.Sonic {
-		withdrawalsHash = types.EmptyWithdrawalsHash
+		withdrawalsHash = &types.EmptyWithdrawalsHash
 	}
 
 	return &EvmHeader{
@@ -102,7 +102,7 @@ func ToEvmHeader(block *inter.Block, index idx.Block, prevHash hash.Event, rules
 		GasUsed:         block.GasUsed,
 		BaseFee:         baseFee,
 		PrevRandao:      prevRandao,
-		WithdrawalsHash: &withdrawalsHash,
+		WithdrawalsHash: withdrawalsHash,
 	}
 }
 
--- a/evmcore/dummy_block.go
+++ b/evmcore/dummy_block.go
@@ -143,28 +143,29 @@ func (h *EvmHeader) EthHeader() *types.Header {
 
 // EvmHeaderJson is simplified version of types.Header, but allowing setting custom hash
 type EvmHeaderJson struct {
-	ParentHash    common.Hash      `json:"parentHash"       gencodec:"required"`
-	UncleHash     common.Hash      `json:"sha3Uncles"       gencodec:"required"`
-	Miner         common.Address   `json:"miner"`
-	Root          common.Hash      `json:"stateRoot"        gencodec:"required"`
-	TxHash        common.Hash      `json:"transactionsRoot" gencodec:"required"`
-	ReceiptHash   common.Hash      `json:"receiptsRoot"     gencodec:"required"`
-	Bloom         types.Bloom      `json:"logsBloom"        gencodec:"required"`
-	Difficulty    *hexutil.Big     `json:"difficulty"       gencodec:"required"`
-	Number        *hexutil.Big     `json:"number"           gencodec:"required"`
-	GasLimit      hexutil.Uint64   `json:"gasLimit"         gencodec:"required"`
-	GasUsed       hexutil.Uint64   `json:"gasUsed"          gencodec:"required"`
-	Time          hexutil.Uint64   `json:"timestamp"        gencodec:"required"`
-	TimeNano      hexutil.Uint64   `json:"timestampNano"`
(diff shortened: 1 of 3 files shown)


LuisPH3:
> NOTE: during the development of this test I realized that by default blocks have a `nil` `Withdrawals` slice, which means that the header's `WithdrawalsHash` will be `nil` as well, and this is generates an error when decoding the field, since it expects a `[32]byte` but instead finds an empty string. Should we make it so that the default value for `Withdrawals` is not `nil`, but rather an empty list ?

This is odd, can you develop. How does the non-sonic go-etherum deal with this? What lines would have to change? How would it be tested ? How does this work for ethereum?

LuisPH3 (dismissed):
Nice concise test, I am missing an enumeration of the properties tested. The comments do not help there neither, they are describing the process but not what is tested (the purpose of running this code).

tests/withdrawals_test.go
```diff
@@ -0,0 +1,64 @@
...
+	// get client and block
+	client, err := net.GetClient()
+	require.NoError(err, "Failed to get the client: %v", err)
+	defer client.Close()
+	block, err := client.BlockByNumber(context.Background(), big.NewInt(0))
+	require.NoError(err, "Failed to get the block: %v", err)
+
+	// check that if no withdrawals are made, then the block withdrawals hash is the empty hash
+	require.Empty(block.Withdrawals())
+	require.Empty(block.Header().WithdrawalsHash)
```
HerbertJordan, line 39: I think the `WithdrawalHash` should be equal to `types.EmptyWithdrawalsHash`, right?

tests/withdrawals_test.go
```diff
@@ -0,0 +1,64 @@
...
+	// decode block
+	stream := rlp.NewStream(buffer, 0)
+	err = newBlock.DecodeRLP(stream)
+	require.NoError(err, "failed to decode block header; %v", err)
+
+	// make empty string hash
+	err = rlp.Encode(buffer, "")
+	require.NoError(err)
+	newHash := common.Hash(crypto.Keccak256(buffer.Bytes()))
+	require.Equal(types.EmptyWithdrawalsHash, newHash)
```
HerbertJordan, line 59: This part looks like it is independent of the rest of the test. It also seems to target the implementation of types and functionality imported from `go-ethereum`.

I guess these steps could be placed in an independent test.

tests/withdrawals_test.go
HerbertJordan, line 1: Something that should be tested:
- fetch a block by its number (like it is done already)
- compute the hash of the block (the block has a member function for that)
- fetch the block by its hash from the client (there if a function for that)
- make sure that this lookup produces the same block information

This makes sure that the block hashing is consistent.

HerbertJordan (changes requested):

tests/withdrawals_test.go
facuMH: I was not aware that this had to be tested as well, but as you put it, it seems easy enough to test.

tests/withdrawals_test.go
facuMH: well, this refers to what I mention in the `NOTE` in the PR description.

tests/withdrawals_test.go
facuMH: yes, this was an independent test, to ensure that we can reproduce `types.EmptyWithdrawalsHash`. Is not needed now, but if in the future we want to reduce dependencies to geth, it was a small step towards that direction

tests/withdrawals_test.go
facuMH: This triggered issue https://github.com/Fantom-foundation/sonic-admin/issues/61 and it will have its own PR

evmcore/dummy_block.go
```diff
@@ -176,24 +177,25 @@ type EvmBlockJson struct {
...
+		UncleHash:       types.EmptyUncleHash,
+		Time:            hexutil.Uint64(h.Time.Unix()),
+		TimeNano:        hexutil.Uint64(h.Time),
+		BaseFee:         (*hexutil.Big)(h.BaseFee),
+		Difficulty:      new(hexutil.Big),
+		PrevRandao:      h.PrevRandao,
+		TotalDiff:       new(hexutil.Big),
+		Hash:            &h.Hash,
+		Epoch:           hexutil.Uint64(hash.Event(h.Hash).Epoch()),
+		WithdrawalsHash: &types.EmptyWithdrawalsHash,
```
HerbertJordan, line 196: I think this field needs to be `nil` if the `Sonic` update is not active, otherwise we are breaking backward compatibility with the Fantom network, which is required for running some of the release tests.

An example of such a test is in line 83:
```
    prevRandao := common.Hash{}
	if rules.Upgrades.Sonic {
		prevRandao.SetBytes([]byte{1}) // TODO provide pseudorandom data?
	}
```

tests/withdrawals_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"math/big"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/stretchr/testify/require"
+)
+
+func TestWithdrawalsCanBeRLPEncodedAndDecoded(t *testing.T) {
```
HerbertJordan, line 15:
```suggestion
func TestWithdrawalFieldsInBlocks(t *testing.T) {
```

tests/withdrawals_test.go
```diff
@@ -0,0 +1,56 @@
...
+	// get client and block
+	client, err := net.GetClient()
+	require.NoError(err, "Failed to get the client: ", err)
+	defer client.Close()
+	block, err := client.BlockByNumber(context.Background(), big.NewInt(0))
+	require.NoError(err, "Failed to get the block: ", err)
+
+	t.Run("verify default values of block's Withdrawals list and hash", func(t *testing.T) {
+		// check that if no withdrawals are made, then the block withdrawals hash is the empty hash
+		require.Equal(types.EmptyWithdrawalsHash, *block.Header().WithdrawalsHash)
```
HerbertJordan, line 37: nit: you need a fresh `require` instance in sub-tests

HerbertJordan (changes requested):
Thanks, the test looks good, but I think we need to preserver backward compatibility. See the comment added to the `dummy_block.go` file.

tests/withdrawals_test.go
```diff
@@ -0,0 +1,57 @@
...
+	requireBase.NoError(err, "Failed to get the block: ", err)
+
+	t.Run("verify default values of block's Withdrawals list and hash", func(t *testing.T) {
+		require := require.New(t)
+		// check that if no withdrawals are made, then the block withdrawals hash is the empty hash
+		require.Equal(types.EmptyWithdrawalsHash, *block.Header().WithdrawalsHash)
+		require.Empty(block.Withdrawals())
+	})
+
+	t.Run("encode and decode works properly", func(t *testing.T) {
```
LuisPH3, line 41: nit: "can be RLP encoded and decoded"

LuisPH3: nit: this case does not need to be nested, can be a free standing function.

tests/withdrawals_test.go
```diff
@@ -0,0 +1,57 @@
...
+	// run endowment to ensure at least one block exists
+	receipt, err := net.EndowAccount(common.Address{42}, 1)
+	requireBase.NoError(err)
+	requireBase.Equal(receipt.Status, types.ReceiptStatusSuccessful, "failed to endow account")
+
+	// get client and block
+	client, err := net.GetClient()
+	requireBase.NoError(err, "Failed to get the client: ", err)
+	defer client.Close()
+	block, err := client.BlockByNumber(context.Background(), receipt.BlockNumber)
```
LuisPH3, line 31: I understand using one client for the complete test scope. Nevertheless, the block is only used in the scope of the next test case and should be queried inside. 

note: i do not think we need cases neither in this test.

tests/withdrawals_test.go
```diff
@@ -0,0 +1,57 @@
...
+	requireBase.NoError(err, "Failed to get the block: ", err)
+
+	t.Run("verify default values of block's Withdrawals list and hash", func(t *testing.T) {
```
LuisPH3, line 34: Nit: withdrawals hash is empty?

tests/withdrawals_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"context"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/stretchr/testify/require"
+)
+
+func TestWithdrawalFieldsInBlocks(t *testing.T) {
```
LuisPH3, line 14: If moving RLP encoding tests out of this function, this test does not need to declare nested cases.

LuisPH3 (commented):
The test checks the property after initialization of the test-net.
I have a question?  Would it be possible to attempt a withdrawal somehow, and check that the blockchain remains having empty hashes afterward?

tests/withdrawals_test.go
facuMH: depending on the revision, for backwards compatibility, it can be nil, but for Sonic it should never be nil, it should have a default value.

tests/withdrawals_test.go
facuMH: RLP encode/decode actually needs the block. in that test case we want to check that blocks coming from the network are healthy to be encoded/decoded. I will change the test case name to reflect that.

tests/withdrawals_test.go
facuMH: sub tests were added simply to distinct what was being checked, since neither affect the context, they could be part of the same test, yes, but I rather leave it as sub test since what is being checked is different

tests/withdrawals_test.go
facuMH: this test case will be renamed to better reflect what is the intent, but will be kept as a sub test.

LuisPH3 (approved):
I understand now the need for doing serialization with a block received from the net.
This is a regression against an observed misconfigured block received in the past.

tests/withdrawals_test.go
```diff
@@ -0,0 +1,71 @@
...
+	requireBase := require.New(t)
```
LuisPH3, line 16: nit: I had this problem as well, where a new `require` object needs to instantiated in a nested scope with a nested testing.T. 
I did not like the solution of renaming the variable because it pollutes the name (it is no longer `expect.Something`) 
the require package implements all the symbols found in the require object in the main namespace as well:
https://pkg.go.dev/github.com/stretchr/testify/require#Equal

By writing `require.NotError(t, err)`  it may get a little more readable in cases where nesting is needed.

facuMH: mmmmm I see, so can probably just instantiate one require and pass the `t` when running it in subtests. I will keep in mind for future work. thanks !!!

evmcore/dummy_block.go
```diff
@@ -85,34 +87,41 @@ func ToEvmHeader(block *inter.Block, index idx.Block, prevHash hash.Event, rules
...
+		Hash:            common.Hash(block.Atropos),
+		ParentHash:      common.Hash(prevHash),
+		Root:            common.Hash(block.Root),
+		Number:          big.NewInt(int64(index)),
+		Time:            block.Time,
+		GasLimit:        math.MaxUint64,
+		GasUsed:         block.GasUsed,
+		BaseFee:         baseFee,
+		PrevRandao:      prevRandao,
+		WithdrawalsHash: &withdrawalsHash,
```
HerbertJordan, line 105: Should this be `nil` if Sonic is not enabled?

gossip/blockproc/evmmodule/evm.go
```diff
@@ -73,17 +73,24 @@ func (p *OperaEVMProcessor) evmBlockWith(txs types.Transactions) *evmcore.EvmBlo
...
+		Hash:            common.Hash(p.block.Atropos),
+		ParentHash:      p.prevBlockHash,
+		Root:            common.Hash{},
+		Time:            p.block.Time,
+		Coinbase:        common.Address{},
+		GasLimit:        math.MaxUint64,
+		GasUsed:         p.gasUsed,
+		BaseFee:         baseFee,
+		PrevRandao:      prevRandao,
+		WithdrawalsHash: &withdrawalsHash,
```
HerbertJordan, line 93: Same question here ... should it be `nil` if Sonic is not enabled?

tests/withdrawals_test.go
```diff
@@ -0,0 +1,71 @@
...
+	requireBase.NoError(err, "Failed to get the client: ", err)
+	defer client.Close()
+
+	t.Run("verify default values of block's Withdrawals list and hash", func(t *testing.T) {
+		require := require.New(t)
+
+		latest, err := client.BlockNumber(context.Background())
+		require.NoError(err, "Failed to get the latest block number: ", err)
+
+		// we check from block 1 onwards because node zero does not consider Sonic Upgrade.
```
HerbertJordan, line 39:
```suggestion
		// we check from block 1 onward because block 0 does not consider Sonic Upgrade.
```

HerbertJordan, line 39: fyi: this is getting fixed by #297

HerbertJordan (dismissed):
Thanks!

gossip/blockproc/evmmodule/evm.go
facuMH: same as the other comment.

tests/withdrawals_test.go
HerbertJordan: Just add a TODO.

HerbertJordan (approved):

PR #283, opened 2024-11-05
Fix race condition in gas-price oracle startup

HerbertJordan:
This PR fixes a race condition in the initialization of the gas-price oracle.

The backend used by the oracle is set as part of the startup procedure of the main sonic server service. However, at this point HTTP service handler may already access the oracle through RPC queries, constituting a race condition.

With this fix, the backend for the gas price oracle is set during construction time.

### DIFF
--- a/gossip/evm_state_reader.go
+++ b/gossip/evm_state_reader.go
@@ -23,14 +23,6 @@ type EvmStateReader struct {
 	gpo   *gasprice.Oracle
 }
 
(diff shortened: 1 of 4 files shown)


LuisPH3 (approved):
I like the elegance of just changing ownerships and not having to add mutexes.

jenikd (approved):
Looks good, thanks

PR #288, opened 2024-11-07
Fix flaky SampleHashes tests

simonlechner:
The testing of the SampleHashes function in the transaction bool was flaky. In both tests a number of expected transaction were sampled in a loop and it was checked that each of the transaction has been returned at least once and at most the number of samplings.
The order of transactions is random, not all transactions have been returned for each test run. This has simply been fixed by increasing the number of times the sampling is called.

Fixes https://github.com/Fantom-foundation/sonic-admin/issues/40

### DIFF
--- a/evmcore/tx_pool_test.go
+++ b/evmcore/tx_pool_test.go
@@ -2647,7 +2647,7 @@ func TestTransactionSlotCount(t *testing.T) {
 	}
 }
 
-func TestSampleHashes(t *testing.T) {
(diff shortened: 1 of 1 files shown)


LuisPH3 (approved):
The test failed because of a probability problem. The proposed solution increases the number of repetitions to reduce the chance of missing expected transactions.

HerbertJordan (approved):

PR #290, opened 2024-11-07
Allow execution of Blob transactions as long as the carry no Blobs

LuisPH3:
This PR adds the modifications needed to allow Blob transactions to proceed into the transaction processor, as long as they do not include Blobs.  

Current implementation would ignore any transaction containing an empty `BlobHashes` field here. Never producing a receipt for such.  [Here](https://github.com/Fantom-foundation/Sonic/blob/09c952eaa6fd6b4b72ed33097f4714275b1e7cd6/evmcore/state_processor.go#L183)

The proposed fix utilizes the same ignore mechanism to drop transactions with blob payload, whilst allowing transactions with empty and nil blob lists.

Blobs are currently being filtered out from submission [here](https://github.com/Fantom-foundation/Sonic/blob/09c952eaa6fd6b4b72ed33097f4714275b1e7cd6/evmcore/tx_pool.go#L661). 
To test the correct execution of blob transaction without blobs, and ignore those which contain blobs in the event of a modified validator which does not reject such; the following was tested:
- Remove pool checks that would prevent the transaction from being part of the next block [here](https://github.com/Fantom-foundation/Sonic/blob/09c952eaa6fd6b4b72ed33097f4714275b1e7cd6/evmcore/tx_pool.go#L661).
- Increase Transaction max size [here](https://github.com/Fantom-foundation/Sonic/blob/09c952eaa6fd6b4b72ed33097f4714275b1e7cd6/evmcore/tx_pool.go#L57).
- Re-rurn the unit test yields the error:
`Error "failed to get transaction receipt: timeout" does not contain "transaction type not supported"`
This means that the transaction was ignored and produced no receipt, although it was accepted in the pool and submitted for a block. 

Part of https://github.com/Fantom-foundation/sonic-admin/issues/16

### DIFF
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -177,6 +177,15 @@ func applyTransaction(
 	// Skip checking of base fee limits for internal transactions.
 	evm.Config.NoBaseFee = msg.SkipAccountChecks
 
+	// For now, Sonic only supports Blob transactions without blob data.
+	if msg.BlobHashes != nil {
+		if len(msg.BlobHashes) > 0 {
+			return nil, 0, true, fmt.Errorf("blob data is not supported")
+		}
+		// PreCheck requires non-nil blobHashes not to be empty
+		msg.BlobHashes = nil
+	}
+
 	// Apply the transaction to the current state (included in the env).
 	result, err := core.ApplyMessage(evm, msg, gp)
 	if err != nil {
--- a/tests/blob_tx_test.go
+++ b/tests/blob_tx_test.go
@@ -2,7 +2,6 @@ package tests
 
 import (
 	"context"
-	"fmt"
 	"testing"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -119,14 +118,10 @@ func createTestBlobTransaction(t *testing.T, ctxt *testContext, data ...[]byte) 
 		copy(blob[:], data)
 
 		blobCommitment, err := kzg4844.BlobToCommitment(&blob)
-		if err != nil {
-			return nil, fmt.Errorf("failed to compute blob commitment: %s", err)
-		}
+		require.NoError(err, "failed to compute blob commitment")
 
 		blobProof, err := kzg4844.ComputeBlobProof(&blob, blobCommitment)
-		if err != nil {
-			return nil, fmt.Errorf("failed to compute blob proof: %s", err)
-		}
+		require.NoError(err, "failed to compute blob proof")
 
 		sidecar.Blobs = append(sidecar.Blobs, blob)
 		sidecar.Commitments = append(sidecar.Commitments, blobCommitment)
--- a/tests/blob_tx_test.go
+++ b/tests/blob_tx_test.go
@@ -1,0 +1,202 @@
+package tests
+
+import (
+	"context"
+	"fmt"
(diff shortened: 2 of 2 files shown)


evmcore/state_processor.go
```diff
@@ -177,6 +177,15 @@ func applyTransaction(
 	// Skip checking of base fee limits for internal transactions.
 	evm.Config.NoBaseFee = msg.SkipAccountChecks
 
+	// For now, Sonic only supports Blob transactions without blob data.
+	if msg.BlobHashes != nil {
```
HerbertJordan, line 181: I think at this point it is to late to filter out the transactions we can not accept. At this point, the list of transactions to be included in a block is already fixed, and with #278 properties will already have been derived from this list of transactions (in particular the order of the transactions).

I think the filtering should happen [here](https://github.com/Fantom-foundation/Sonic/blob/a75bcb23ff6345f22dbbb53fe61382d583354c36/gossip/c_block_callbacks.go#L311-L313), before the list of transactions of a block is passed to the processor. Also at this place, the transaction re-ordering is going to be added.

HerbertJordan (changes requested):
Test looks good, but the filter preventing the execution of the blob transactions should be moved to an earlier point in the pipeline (see comment).

evmcore/state_processor.go
LuisPH3: Yes, this is what we once discussed. Nevertheless, I discovered that it does exist the possibility that the processor skips the transaction (what was happening already with the ones with non-nil but empty blobhashes).

questions: 
- shall it continue to be ignored or do we want to craft a failure receipt? 
- the patching of empty->nil blobhases still needs to be done at the current place. Unless we want to reject transactions with non-nil but empty bloblhases list.

HerbertJordan: You are right, we do already skip other transactions as well. So my original thought of being able to verify the scrambling is anyway already broken -- and as I just realized, also not necessary. So yes, in this case we can leave it as you suggested it.

HerbertJordan (approved):

tests/blob_tx_test.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	for _, data := range data {
+		var blob kzg4844.Blob // Define a blob array to hold the large data payload, blobs are 128kb in length
+		copy(blob[:], data)
+
+		blobCommitment, err := kzg4844.BlobToCommitment(&blob)
+		if err != nil {
+			return nil, fmt.Errorf("failed to compute blob commitment: %s", err)
+		}
```
facuMH, line 124:
```suggestion
		require.NoError(err, "failed to compute blob commitment: ", err)
```

tests/blob_tx_test.go
```diff
@@ -0,0 +1,202 @@
...
+
+		blobProof, err := kzg4844.ComputeBlobProof(&blob, blobCommitment)
+		if err != nil {
```
facuMH, line 127: require.NoError

facuMH (approved):
looks good to me. I like the testing desing, just a couple of small comments

facuMH (approved):

HerbertJordan (approved):

PR #294, opened 2024-11-11
Modify RPC function GetBlockReceipts to accept block hash as parameter

jenikd:
This PR adds possibility to use not only a block numer, but also  block hash as a parameter for this RPC function

Refers to https://github.com/Fantom-foundation/sonic-admin/issues/53


HerbertJordan (approved):
Looks good, thanks.

PR #295, opened 2024-11-11
Segregate integration tests into different processes.

LuisPH3:
This PR segregates the different integration tests into their own package names. Therefore, they will be compiled into separate binaries which `go test` will execute in parallel. 
This is done to speedup execution of unit tests, which during race detection can easily exceed timeout settings. 

In addition:
- Count and test-net-test reuse running integration networks to avoid overhead
- Timeout time for the integration network startup has been increased to 5 minutes. Because race-enabled tests would time-out during initialization of the network, and produce false positives. 

This pr is part of https://github.com/Fantom-foundation/sonic-admin/issues/50
This pr helps towards https://github.com/Fantom-foundation/sonic-admin/issues/47.

### DIFF
--- a/tests/counter/counter_test.go
+++ b/tests/counter/counter_test.go
@@ -1,23 +1,24 @@
-package tests
+package counter_test
 
 import (
 	"math"
 	"math/big"
 	"testing"
 
+	"github.com/Fantom-foundation/go-opera/tests"
 	"github.com/Fantom-foundation/go-opera/tests/contracts/counter"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 )
 
 func TestCounter_CanIncrementAndReadCounterFromHead(t *testing.T) {
-	net, err := StartIntegrationTestNet(t.TempDir())
+	net, err := tests.StartIntegrationTestNet(t.TempDir())
 	if err != nil {
 		t.Fatalf("Failed to start the fake network: %v", err)
 	}
 	defer net.Stop()
 
 	// Deploy the counter contract.
-	contract, _, err := DeployContract(net, counter.DeployCounter)
+	contract, _, err := tests.DeployContract(net, counter.DeployCounter)
 	if err != nil {
 		t.Fatalf("failed to deploy contract; %v", err)
 	}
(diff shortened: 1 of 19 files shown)


tests/counter/counter_test.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	defer net.Stop()
+
+	t.Run("can increment and read counter from head", func(t *testing.T) {
+
+		// Deploy the counter contract.
+		contract, _, err := tests.DeployContract(net, counter.DeployCounter)
+		if err != nil {
+			t.Fatalf("failed to deploy contract; %v", err)
+		}
```
facuMH, line 26: why not do it before the `t.Run` and use `contract` and `receipt` for both test cases ?

tests/integration_test_net.go
```diff
@@ -44,54 +47,112 @@ import (
...
+	if err != nil {
+		return false
+	}
+	listener.Close()
+	return true
+}
+
+func getFreePort() (int, error) {
+	var port int
+	for i := 0; i < 10; i++ {
```
facuMH, line 67: should `10` be a constant with a more representative name, or even a configurable parameter ?

tests/counter/counter_test.go
```diff
@@ -0,0 +1,91 @@
...
+			counter, err := contract.GetCount(nil)
+			if err != nil {
+				t.Fatalf("failed to get counter value; %v", err)
+			}
+
+			if counter.Cmp(new(big.Int).SetInt64(int64(i))) != 0 {
+				t.Fatalf("unexpected counter value; expected %d, got %v", i, counter)
+			}
+
+			_, err = net.Apply(contract.IncrementCounter)
```
facuMH, line 39: would it make sense to check the `receipt.Status` for a success ?

tests/counter/counter_test.go
LuisPH3: Because then, changes done in the account would leak to the other test. This way the test is stateless.

tests/counter/counter_test.go
LuisPH3: I did not want to rewrite the test, just superficial rearrangement. 
but yes. This should be added.

tests/integration_test_net.go
LuisPH3: let me see what can we do, this technique has been used in norma already.

LuisPH3:
This PR has been moved back to draft because of a recurring failure in CI:

When running, `go test ./tests/...` everything seems to be alright. But when running `go test ./...` there are spurious test failures and/or termination signal received. This makes me think that there is yet some shared resource that does not get randomized correctly. 


Because the race detector did not find any new race, when running it locally. I will park this work for the time being and focus on other tasks.
I will come back to this issue once the other issues are dispatched.

LuisPH3:
Although the segregation is feasible, memory consumption makes it unpractical. This PR is closed and development can be continued if memory usage is dramatically reduced.

LuisPH3:
Although the segregation is feasible, memory consumption makes it unpractical. This PR is closed and development can be continued if memory usage is dramatically reduced.

PR #297, opened 2024-11-11
Integration of Sonic's per-block base-fee adjustment

HerbertJordan:
This PR integrates Sonic's base-fee adjustment algorithm into the Sonic client. After this PR, base-fees are updated with each block and no longer determined at the epoch boundary.

To integrate this feature, two additional values need to be stored in blocks:
- the nano-second part of the block **time**, using 4 bytes
- the **block duration**, defined as the duration in nanoseconds between the a block's predecessor and the block itself (8 bytes)

Both values are required for the computation of base-fee prices. Since fields can not be arbitrarily added to blocks, a pre-existing, previously unused 32-byte "extra-data" fields was used. The two new values are stored in big endian format in the first 12 bytes of this value.

Besides these changes, this PR also fixes the handling of the `GasLimit` field in blocks, which previously was filled with `math.MaxUint64` but presented to RPC clients as the 6-byte value `0xffffffffffff` presumably in order to avoid parsing or overflow errors for client code using signed 64-bit integers. With this change, an actual value -- the `rules.Blocks.MaxBlockGas` of the network configuration rules -- is used as the Gas limit per block.

### DIFF
--- a/gossip/apply_genesis.go
+++ b/gossip/apply_genesis.go
@@ -4,8 +4,10 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"time"
 
 	"github.com/Fantom-foundation/go-opera/evmcore"
+	"github.com/Fantom-foundation/go-opera/gossip/gasprice"
 	"github.com/Fantom-foundation/go-opera/inter"
 	"github.com/Fantom-foundation/go-opera/inter/iblockproc"
 	"github.com/Fantom-foundation/go-opera/inter/ibr"
@@ -51,6 +53,9 @@ func (s *Store) ApplyGenesis(g genesis.Genesis) (err error) {
 	s.SetBlockEpochState(topEr.BlockState, topEr.EpochState)
 	s.FlushBlockEpochState()
 
+	s.SetGenesisID(g.GenesisID)
+	s.SetGenesisBlockIndex(topEr.BlockState.LastBlock.Idx)
+
 	// write blocks
 	rules := s.GetRules()
 	gasLimit := rules.Blocks.MaxBlockGas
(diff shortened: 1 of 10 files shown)


gossip/apply_genesis.go
```diff
@@ -60,15 +65,28 @@ func (s *Store) ApplyGenesis(g genesis.Genesis) (err error) {
 		WithTime(evmcore.FakeGenesisTime-1). // TODO: extend genesis generator to provide time
 		WithGasLimit(gasLimit).
 		WithStateRoot(common.Hash{}). // TODO: get proper state root from genesis data
-		WithBaseFee(big.NewInt(0)).   // TODO: set initial base fee according to the rules
+		WithBaseFee(gasprice.GetInitialBaseFee(rules.Economy)).
```
LuisPH3, line 68: I think this would solve some problems I had with initial price testing.

gossip/c_block_callbacks.go
```diff
@@ -380,9 +384,16 @@ func consensusCallbackBeginBlockFn(
...
-						"age", utils.PrettyDuration(blockAge), "t", utils.PrettyDuration(now.Sub(start)))
+					log.Info("New block",
+						"index", blockCtx.Idx,
+						"id", block.Hash(),
+						"gas_used", evmBlock.GasUsed,
+						"gas_rate", float64(evmBlock.GasUsed)/blockDuration.Seconds(),
+						"base_fee", evmBlock.BaseFee.String(),
+						"txs", fmt.Sprintf("%d/%d", len(evmBlock.Transactions), len(skippedTxs)),
+						"age", utils.PrettyDuration(blockAge),
+						"t", utils.PrettyDuration(now.Sub(start)),
```
LuisPH3, line 395: This start value is misleading, we may want to remove it.
It would be more interesting to have the blockDuratin and the blockctx.time here. 
After all, the timestamp is introduced by the logging itself.

LuisPH3 (approved):

gossip/c_block_callbacks.go
HerbertJordan: This log is only partially targeted for human consumption. It is, unfortunately, also used by Norma to collect monitoring data from the client. The fields, as they have been used before and the ones added, are needed in Norma to plot base-fee prices and throughput rates over time.

jenikd (commented):
Looks good, thanks for these changes. Having correct value for GasLimit is great.

PR #298, opened 2024-11-12
Add configurable structured logs limit for RPC debug

jenikd:
This PR introduces configurable limit for number of logs, created during debug of EVM with debug namespace
Parameter:
- `rpc.structloglimit `

Default set to 2000 logs

### DIFF
--- a/config/flags/flags.go
+++ b/config/flags/flags.go
@@ -309,7 +309,7 @@ var (
 	}
 	StructLogLimitFlag = cli.IntFlag{
 		Name:  "rpc.structloglimit",
-		Usage: "Limit maximum number of debug logs for structured EVM logs",
+		Usage: "Limit maximum number of debug logs for structured EVM logs, 0=unlimited, negative value means no log results",
 		Value: gossip.DefaultConfig(cachescale.Identity).StructLogLimit,
 	}
 	ModeFlag = cli.StringFlag{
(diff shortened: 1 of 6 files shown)


config/flags/flags.go
```diff
@@ -307,6 +307,11 @@ var (
 		Usage: "Limit maximum size in some RPC calls execution",
 		Value: gossip.DefaultConfig(cachescale.Identity).MaxResponseSize,
 	}
+	StructLogLimitFlag = cli.IntFlag{
```
HerbertJordan, line 310: Looks like this flag is not used anywhere ... I guess it should be integrated somewhere in a list of flags and read?

config/flags/flags.go
```diff
@@ -307,6 +307,11 @@ var (
...
+		Name:  "rpc.structloglimit",
+		Usage: "Limit maximum number of debug logs for structured EVM logs",
```
HerbertJordan, line 312: Can you add something about the interpretation of 0 and negative values?

HerbertJordan (changes requested):

config/flags/flags.go
jenikd: Done

config/flags/flags.go
jenikd: You are right, thanks for noticing

HerbertJordan (approved):

PR #301, opened 2024-11-13
Use block-header hashes to address blocks

HerbertJordan:
This PR aligns the block format used by Sonic with the format defined by Ethereum. In particular, hashes to identify blocks are now equivalent to the hashes of the respective block headers.

**Note:** this PR breaks backward compatibility with the Fantom mainnet. After this change, genesis data of former Fantom networks can no longer be used and resulting block hashes are incompatible.

### Changes

The main change required for this is the re-design of the storage format of blocks as they are retained on the disk. The previous format was a summary of the consensus events that formed to block, in particular including the list of `Events` the block was created from, as well as the Atropos and various lists of transactions and flags composing the block. When block information was required, it was derived from the available set on demand. This information was incomplete.

The new format stores block information explicitly on the disk. Everything required to reproduce the block (header) in its full detail is captured, to facilitate a complete reproduction. In particular, the following fields got added:
- ParentHash ... the hash of the parent block
- Difficulty ... the difficulty value for this block
- GasLimit ... the maximum amount of gas allowed for this block
- BaseFee ... the minimum gas price for this block
- PrevRandao ... the random value generated by the consensus for this block
- ReceiptsHashRoot ... cryptographic summary of all receipts in the block
- LogBloom ... summary of all log messages in the block

To simplify code creating blocks, a `BlockBuilder` was added. This allows to build blocks gradually -- by adding pairs of transactions and receipts -- and compute derived values in a final `Build` step.

Finally, the creation of *genesis* blocks was adjusted to satisfy the requirements of well defined blocks (in particular to use correct parent pointers). A major change there is that Block 0, which was previously retained implicit through hard-coded values in the code is now explicitly restored in the block store.

### Testing

As this targets a core element of the system, most preexisting tests cover are affected by this change. In particular integration tests in the `./tests` directory running fake nets are covering the creation, storage, and retrieval of block information. All of those tests pass without any changes.

Additionally, a new integration test suite, `./tests/block_header_test.go`, is added by this PR, checking the following invariants:
- block headers retrieved for block height x are indeed for block x
- the parent hash of a block corresponds to the hash of the header of the block's predecessor
- the gas usage of every block is at most the the gas limit of the block

The test suite provides a general framework for defining such block invariant properties and tests. A list of future tests to be added is left as a TODO for a follow-up PR.

Finally, this change was tested using Norma to verify through a smoke test that the resulting client can handle multi-node settups.

### For Reviewers

The suggested way to review this PR is in the following order:
- start with reviewing the changes in the `inter` package. This is the core of the change, defining the new storage format and hashing schema
- continue with `tests/block_header_test.go`, which defines the main objective of this change; in particular, it contains a test making sure that parent-references of blocks correspond to the hash of the parent's header
- review `gossip/c_block_callbacks.go` which contains the changes required for the block creation; related to this is the change `gossip/blockproc/evmmodule/evm.go` altering the list of produced receipts to include `nil` values for skipped transactions
- review changes in `gossip/c_llr_callbacks.go` which cover adaptations for the processing of blocks during genesis imports such that also those blocks satisfy block-invariants 
- the rest of the changes are adaptations to those changes above

### DIFF
--- a/cmd/sonictool/check/live.go
+++ b/cmd/sonictool/check/live.go
@@ -8,6 +8,7 @@ import (
 	"github.com/Fantom-foundation/Carmen/go/database/mpt"
 	"github.com/Fantom-foundation/Carmen/go/database/mpt/io"
 	carmen "github.com/Fantom-foundation/Carmen/go/state"
+	"github.com/Fantom-foundation/lachesis-base/hash"
 	"github.com/Fantom-foundation/lachesis-base/utils/cachescale"
 	"github.com/ethereum/go-ethereum/log"
 )
@@ -44,7 +45,7 @@ func checkLiveBlockRoot(dataDir string, cacheRatio cachescale.Func) error {
 	if lastBlock == nil {
 		return fmt.Errorf("verification failed - unable to get the last block (%d) from gdb", lastBlockIdx)
 	}
-	err = gdb.EvmStore().CheckLiveStateHash(lastBlockIdx, lastBlock.Root)
+	err = gdb.EvmStore().CheckLiveStateHash(lastBlockIdx, hash.Hash(lastBlock.StateRoot))
 	if err != nil {
 		return fmt.Errorf("checking live state failed: %w", err)
 	}
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -799,7 +799,7 @@ func (s *PublicBlockChainAPI) GetHeaderByHash(ctx context.Context, hash common.H
 
 func (s *PublicBlockChainAPI) getBlockReceipts(ctx context.Context, blkNumber rpc.BlockNumber) (types.Receipts, error) {
 	if blkNumber == rpc.EarliestBlockNumber {
-		return nil, nil
+		return types.Receipts{}, nil
 	}
 	return s.b.GetReceiptsByNumber(ctx, blkNumber)
 }
--- a/gossip/apply_genesis.go
+++ b/gossip/apply_genesis.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"math/big"
 
+	"github.com/Fantom-foundation/go-opera/evmcore"
+	"github.com/Fantom-foundation/go-opera/inter"
 	"github.com/Fantom-foundation/go-opera/inter/iblockproc"
 	"github.com/Fantom-foundation/go-opera/inter/ibr"
 	"github.com/Fantom-foundation/go-opera/inter/ier"
@@ -50,11 +52,23 @@ func (s *Store) ApplyGenesis(g genesis.Genesis) (err error) {
 	s.FlushBlockEpochState()
 
 	// write blocks
+	rules := s.GetRules()
+	gasLimit := rules.Blocks.MaxBlockGas
+
+	s.SetBlock(0, inter.NewBlockBuilder().
+		SetNumber(0).
+		SetTime(evmcore.FakeGenesisTime-1). // TODO: extend genesis generator to provide time
+		SetGasLimit(gasLimit).
+		SetStateRoot(common.Hash{}). // TODO: get proper has from genesis data
+		SetBaseFee(big.NewInt(0)).
+		Build(),
+	)
+
 	baseFee := prevEs.Rules.Economy.MinGasPrice
 	blobGasPrice := big.NewInt(1) // TODO issue #147
 	var lastBlock ibr.LlrIdxFullBlockRecord
 	g.Blocks.ForEach(func(br ibr.LlrIdxFullBlockRecord) bool {
-		s.WriteFullBlockRecord(baseFee, blobGasPrice, br)
+		s.WriteFullBlockRecord(baseFee, blobGasPrice, gasLimit, br)
 		if br.Idx > lastBlock.Idx {
 			lastBlock = br
 		}
--- a/gossip/apply_genesis.go
+++ b/gossip/apply_genesis.go
@@ -95,7 +95,7 @@ func (s *Store) ApplyGenesis(g genesis.Genesis) (err error) {
 		}
 	} else { // no S5 section in the genesis file
 		// Import legacy EVM genesis section
-		err = s.evm.ImportLegacyEvmData(g.RawEvmItems, uint64(lastBlock.Idx), common.Hash(lastBlock.Root))
+		err = s.evm.ImportLegacyEvmData(g.RawEvmItems, uint64(lastBlock.Idx), common.Hash(lastBlock.StateRoot))
 		if err != nil {
 			return fmt.Errorf("import of legacy genesis data into StateDB failed; %v", err)
 		}
@@ -104,10 +104,10 @@ func (s *Store) ApplyGenesis(g genesis.Genesis) (err error) {
 	if err := s.evm.Open(); err != nil {
 		return fmt.Errorf("unable to open EvmStore to check imported state: %w", err)
 	}
-	if err := s.evm.CheckLiveStateHash(lastBlock.Idx, lastBlock.Root); err != nil {
+	if err := s.evm.CheckLiveStateHash(lastBlock.Idx, lastBlock.StateRoot); err != nil {
 		return fmt.Errorf("checking imported live state failed: %w", err)
 	} else {
-		s.Log.Info("StateDB imported successfully, stateRoot matches", "index", lastBlock.Idx, "root", lastBlock.Root)
+		s.Log.Info("StateDB imported successfully, stateRoot matches", "index", lastBlock.Idx, "root", lastBlock.StateRoot)
 	}
 
 	s.SetGenesisID(g.GenesisID)
--- a/gossip/apply_genesis.go
+++ b/gossip/apply_genesis.go
@@ -56,11 +56,11 @@ func (s *Store) ApplyGenesis(g genesis.Genesis) (err error) {
 	gasLimit := rules.Blocks.MaxBlockGas
 
 	s.SetBlock(0, inter.NewBlockBuilder().
-		SetNumber(0).
-		SetTime(evmcore.FakeGenesisTime-1). // TODO: extend genesis generator to provide time
-		SetGasLimit(gasLimit).
-		SetStateRoot(common.Hash{}). // TODO: get proper state root from genesis data
-		SetBaseFee(big.NewInt(0)).   // TODO: set initial base fee according to the rules
+		WithNumber(0).
+		WithTime(evmcore.FakeGenesisTime-1). // TODO: extend genesis generator to provide time
+		WithGasLimit(gasLimit).
+		WithStateRoot(common.Hash{}). // TODO: get proper state root from genesis data
+		WithBaseFee(big.NewInt(0)).   // TODO: set initial base fee according to the rules
 		Build(),
 	)
 
--- a/gossip/apply_genesis.go
+++ b/gossip/apply_genesis.go
@@ -59,8 +59,8 @@ func (s *Store) ApplyGenesis(g genesis.Genesis) (err error) {
 		SetNumber(0).
 		SetTime(evmcore.FakeGenesisTime-1). // TODO: extend genesis generator to provide time
 		SetGasLimit(gasLimit).
-		SetStateRoot(common.Hash{}). // TODO: get proper has from genesis data
-		SetBaseFee(big.NewInt(0)).
+		SetStateRoot(common.Hash{}). // TODO: get proper state root from genesis data
+		SetBaseFee(big.NewInt(0)).   // TODO: set initial base fee according to the rules
 		Build(),
 	)
 
--- a/gossip/blockproc/drivermodule/driver_txs.go
+++ b/gossip/blockproc/drivermodule/driver_txs.go
@@ -66,7 +66,7 @@ func InternalTxBuilder(statedb state.StateDB) func(calldata []byte, addr common.
 		if nonce == math.MaxUint64 {
 			nonce = statedb.GetNonce(common.Address{})
 		}
-		tx := types.NewTransaction(nonce, addr, common.Big0, 1e10, common.Big0, calldata)
+		tx := types.NewTransaction(nonce, addr, common.Big0, 1e6, common.Big0, calldata)
 		nonce++
 		return tx
 	}
--- a/gossip/c_block_callbacks.go
+++ b/gossip/c_block_callbacks.go
@@ -333,9 +333,6 @@ func consensusCallbackBeginBlockFn(
 							}
 						}
 					}
-					for _, tx := range append(preInternalTxs, internalTxs...) {
-						store.evm.SetTx(tx.Hash(), tx)
-					}
 
 					bs.LastBlock = blockCtx
 					bs.CheatersWritten = uint32(bs.EpochCheaters.Len())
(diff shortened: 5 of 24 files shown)


inter/block.go
```diff
@@ -1,45 +1,188 @@
...
+	"github.com/Fantom-foundation/lachesis-base/inter/idx"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/trie"
 )
 
+// Block represents the on-disk storage format of a block. It contains all
+// fields required to reconstruct the block header, as well as a list of
+// hashes of the transactions been executed as part of the represented block.
```
facuMH, line 16:
```suggestion
// hashes of the transactions being executed as part of the represented block.
```

tests/block_header_test.go
```diff
@@ -0,0 +1,91 @@
...
+	net, err := StartIntegrationTestNet(t.TempDir())
+	require.NoError(err)
+	defer net.Stop()
+
+	// Produce a few blocks on the network.
+	for range numBlocks {
+		_, err := net.EndowAccount(common.Address{42}, 100)
+		if err != nil {
+			t.Fatalf("failed to endow account; %v", err)
+		}
```
facuMH, line 26:
```suggestion
		require.NoError(err,"failed to endow account; %v")
```

tests/block_header_test.go
```diff
@@ -0,0 +1,91 @@
...
+	// - difficulty and nonce is set to 0
+	// - time is progressing strictly monotonically and approximately matches the current time
+	// - the random mixDigest field is different for each block
+}
+
+func testHeaders_NumberMatchesPosition(t *testing.T, headers []*types.Header) {
+	require := require.New(t)
+	for i, header := range headers {
+		require.Equal(header.Number.Uint64(), uint64(i))
+	}
```
facuMH, line 70: if I understood correctly from documentation you can also do 
```suggestion
	for i, header := range headers {
		require.Equal(t, header.Number.Uint64(), uint64(i))
	}
```
no need to define a new instance of `require`, but it would use the provided `t`

facuMH: same for tests below

tests/integration_test_net.go
```diff
@@ -189,7 +189,7 @@ func (n *IntegrationTestNet) GetReceipt(txHash common.Hash) (*types.Receipt, err
 	const maxDelay = 100 * time.Millisecond
 	now := time.Now()
 	delay := time.Millisecond
-	for time.Since(now) < 10*time.Second {
+	for time.Since(now) < 100*time.Second {
```
facuMH, line 192: what's the motivation for this increase ?

inter/block.go
facuMH, line 16:
```suggestion
// hashes of the transactions being executed as part of the represented block.
```

tests/block_header_test.go
facuMH, line 26: use require

tests/block_header_test.go
facuMH, line 70: IRRC from the documentation of require, you can do this instead of a new instance 
```suggestion
	for i, header := range headers {
		require.Equal(t, header.Number.Uint64(), uint64(i))
	}
```
and it will use the provided `t`.

also in the test below

tests/integration_test_net.go
facuMH, line 192: what is the motivation for this increase in time ?

gossip/c_block_callbacks.go
```diff
@@ -178,11 +178,20 @@ func consensusCallbackBeginBlockFn(
...
+
+				maxBlockGas := es.Rules.Blocks.MaxBlockGas
+
+				blockBuilder := inter.NewBlockBuilder().
+					SetEpoch(es.Epoch).
+					SetNumber(number).
+					SetParentHash(lastBlockHeader.Hash).
+					SetTime(blockCtx.Time).
+					SetPrevRandao(prevRandao).
+					SetGasLimit(maxBlockGas)
```
facuMH, line 194: nice

cmd/sonictool/check/live.go
```diff
@@ -44,7 +45,7 @@ func checkLiveBlockRoot(dataDir string, cacheRatio cachescale.Func) error {
 	if lastBlock == nil {
 		return fmt.Errorf("verification failed - unable to get the last block (%d) from gdb", lastBlockIdx)
 	}
-	err = gdb.EvmStore().CheckLiveStateHash(lastBlockIdx, lastBlock.Root)
+	err = gdb.EvmStore().CheckLiveStateHash(lastBlockIdx, hash.Hash(lastBlock.StateRoot))
```
facuMH, line 48: future work: should we try to unify hash types ?

ethapi/api.go
```diff
@@ -799,7 +799,7 @@ func (s *PublicBlockChainAPI) GetHeaderByHash(ctx context.Context, hash common.H
 
 func (s *PublicBlockChainAPI) getBlockReceipts(ctx context.Context, blkNumber rpc.BlockNumber) (types.Receipts, error) {
 	if blkNumber == rpc.EarliestBlockNumber {
-		return nil, nil
+		return types.Receipts{}, nil
```
facuMH, line 802: is this because `nil` would be interpreted as `skipped`?

gossip/apply_genesis.go
```diff
@@ -50,11 +52,23 @@ func (s *Store) ApplyGenesis(g genesis.Genesis) (err error) {
...
 
 	// write blocks
+	rules := s.GetRules()
+	gasLimit := rules.Blocks.MaxBlockGas
+
+	s.SetBlock(0, inter.NewBlockBuilder().
+		SetNumber(0).
+		SetTime(evmcore.FakeGenesisTime-1). // TODO: extend genesis generator to provide time
+		SetGasLimit(gasLimit).
+		SetStateRoot(common.Hash{}). // TODO: get proper has from genesis data
```
facuMH, line 62:
```suggestion
		SetStateRoot(common.Hash{}). // TODO: get proper hash from genesis data
```

gossip/evm_state_reader.go
```diff
@@ -111,20 +114,28 @@ func (r *EvmStateReader) getBlock(h hash.Event, n idx.Block, readTxs bool) *evmc
...
 	es := r.store.GetHistoryEpochState(epoch)
 	var rules opera.Rules
 	if es != nil {
 		rules = es.Rules
 	}
-	var prev hash.Event
+
+	// There is no epoch state for epoch 0 comprising block 0.
+	// For this epoch, London and Sonic upgrades are enabled.
+	if epoch == 0 {
```
facuMH, line 126: was this change not also included in another PR ? I think it's a good change, just making sure its inclusion here is intended.

facuMH (approved):
great work !! just a couple of superficial comments.

inter/block.go
kjezek, line 16:
```suggestion
// hashes of the transactions being executed as part of the represented block.
```

inter/block.go
```diff
@@ -1,45 +1,188 @@
...
+	Number               uint64
+	ParentHash           common.Hash
+	StateRoot            common.Hash
+	Time                 Timestamp
+	Difficulty           uint64
+	GasLimit             uint64
+	GasUsed              uint64
+	BaseFee              *big.Int
+	PrevRandao           common.Hash
+	TransactionsHashRoot common.Hash
```
LuisPH3, line 36: what is the difference with transactionHashes? 
is this hash redundant when considering the parentHash and the transactionHashes?

inter/block.go
```diff
@@ -1,45 +1,188 @@
...
-		if skipCount < len(skippedTxs) && skippedTxs[skipCount] == uint32(i) {
-			skipCount++
-		} else {
-			filteredTxs = append(filteredTxs, tx)
-		}
+func NewBlockBuilder() *BlockBuilder {
+	return &BlockBuilder{}
+}
+
+func (b *BlockBuilder) SetNumber(number uint64) *BlockBuilder {
```
LuisPH3, line 113: nit: builder pattern usually writes "with" instead of "set"

kjezek (approved):
as far as I can see it looks good. Thanks

tests/block_header_test.go
LuisPH3:
```suggestion
		if err != nil {
			t.Fatalf("failed to endow account")
		}
```

tests/integration_test_net.go
```diff
@@ -91,7 +91,7 @@ func StartIntegrationTestNet(directory string) (*IntegrationTestNet, error) {
 	}
 	defer client.Close()
 
-	const timeout = 30 * time.Second
+	const timeout = 300 * time.Second
```
LuisPH3, line 94: This leaked here, but I think it is harmless and you obviously need it.

tests/block_header_test.go
```diff
@@ -0,0 +1,91 @@
...
+	require.GreaterOrEqual(lastBlock.NumberU64(), uint64(numBlocks))
+
+	headers := []*types.Header{}
+	for i := int64(0); i < int64(lastBlock.NumberU64()); i++ {
+		header, err := client.HeaderByNumber(context.Background(), big.NewInt(i))
+		require.NoError(err)
+		headers = append(headers, header)
+	}
+
+	t.Run("NumberMatchesPosition", func(t *testing.T) {
```
LuisPH3, line 44:
```suggestion
	t.Run("BlockNumberEqualsPositionInChain", func(t *testing.T) {
```

tests/block_header_test.go
```diff
@@ -0,0 +1,91 @@
...
+	require := require.New(t)
+	for i, header := range headers {
+		require.Equal(header.Number.Uint64(), uint64(i))
+	}
+}
+
+func testHeaders_ParentHashCoversParentContent(t *testing.T, headers []*types.Header) {
+	require := require.New(t)
+
+	// All other blocks have a parent hash that matches the previous block's hash.
```
LuisPH3, line 76:
```suggestion
	// Every block except for the first one have a parent hash that matches the previous block's hash.
```

gossip/c_block_callbacks.go
```diff
@@ -178,11 +178,20 @@ func consensusCallbackBeginBlockFn(
...
-					Events:  hash.Events(confirmedEvents),
-				}
+				prevRandao := computePrevRandao(confirmedEvents)
+
+				number := uint64(blockCtx.Idx)
+				lastBlockHeader := evmStateReader.GetHeaderByNumber(number - 1)
+
+				maxBlockGas := es.Rules.Blocks.MaxBlockGas
+
+				blockBuilder := inter.NewBlockBuilder().
```
LuisPH3, line 188: Ideally, the builder should be created inside the blockFn lambda, to avoid state pollution in case this function is patched in the future.

LuisPH3 (approved):
The PR replaces the block hashing algorithm and introduces infrastructure to facilitate the task.
- I do not see problems with the hashed data.
- The builder is keep alive for a little too long, being shared with a background thread. I understand this is a limitation of the current pipeline, but ideally the builder should not leak state into other routines to reduce chances for state pollution.

inter/ibr/inter_block_records.go
```diff
@@ -10,21 +10,21 @@ import (
...
-	Receipts []*types.ReceiptForStorage
-	Time     inter.Timestamp
-	GasUsed  uint64
+	BlockHash hash.Hash
+	StateRoot hash.Hash
+	Txs       types.Transactions
+	Receipts  []*types.ReceiptForStorage
+	Time      inter.Timestamp
+	GasUsed   uint64
 }
```
thaarok, line 28: Note: LlrBlockVote/LlrFullBlockRecord structs are relevant only for legacy Opera mainnet events support.
If we would break their compatibility (the changes above does not), they can be removed immediately.

inter/block.go
```diff
@@ -1,45 +1,188 @@
...
+		b.hash = b.GetEthereumHeader().Hash()
+	}
+	return b.hash
+}
+
+// uncleHash is the hash to be used for the uncle field in Ethereum headers if
+// there are no uncles. See https://eips.ethereum.org/EIPS/eip-4844.
+var uncleHash = common.BytesToHash(hexutil.MustDecode(
+	"0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
+))
```
thaarok, line 62: You can use `types.EmptyUncleHash` defined in go-ethereum. (the dependency is already here through the `types.Header` type)

gossip/evm_state_reader.go
thaarok: Should not we include the epoch 0 state into the genesis file instead, like it is already done in this PR with the block 0?

thaarok (approved):
It is a courageous change, but I don't see any issue.

tests/integration_test_net.go
HerbertJordan: I have fixed this so many times manually when debugging, and it does not hurt when everything is working fine, so I decided to include it in this PR.

With this increase you have enough time to step through code with the debugger, without risking that you run into a timeout.

cmd/sonictool/check/live.go
HerbertJordan: Very future work, yes. Likely not in this codebase.

ethapi/api.go
HerbertJordan: The RPC handler consuming this make a difference between an empty list of receipts and the `nil` list when encoding the receipts-root-hash into a header. A `nil` list results in a 0-root hash, while a empty list results in the hash of an empty list.

This is not related to skipped transactions.

inter/block.go
HerbertJordan: The `TransactionHashes` are a list of of hashes of the transactions being part of a given block. They are used to reference transactions being stored in a different table in case the full block needs to be restored.

The `TransactionsHashRoot` is an aggregated hash over all transactions that requires the list of all (full) transactions to be recomputed. It could be recomputed based on the list of `TransactionHashes` if there is access to the transaction table.

When requesting the header of a block, only the `TransactionHashRoot` is required, as it commits all contained transactions and their order. I think it is a sensible improvement to store the root explicitly for faster access.

inter/block.go
HerbertJordan: updated

tests/block_header_test.go
HerbertJordan: Ok ... but I do not think this improves readability, so I would stick with it as it is.

gossip/evm_state_reader.go
HerbertJordan: I tried getting this to work, but I could not.

I think the genesis generation process requires a general check-up. I created an issue for that and added a comment to the code that this requires a cleanup ([#72](https://github.com/Fantom-foundation/sonic-admin/issues/72)).

gossip/blockproc/drivermodule/driver_txs.go
```diff
@@ -66,7 +66,7 @@ func InternalTxBuilder(statedb state.StateDB) func(calldata []byte, addr common.
 		if nonce == math.MaxUint64 {
 			nonce = statedb.GetNonce(common.Address{})
 		}
-		tx := types.NewTransaction(nonce, addr, common.Big0, 1e10, common.Big0, calldata)
+		tx := types.NewTransaction(nonce, addr, common.Big0, 1e6, common.Big0, calldata)
```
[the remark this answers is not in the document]
HerbertJordan: @jmpike what would be a good upper boundary for those inner transactions?

LuisPH3 (approved):

facuMH (approved):

thaarok (approved):

gossip/blockproc/drivermodule/driver_txs.go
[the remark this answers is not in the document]
HerbertJordan: I created a follow-up issue for this to look into this independently: [#73](https://github.com/Fantom-foundation/sonic-admin/issues/73)

PR #303, opened 2024-11-14
Add an integration test covering the restart of a node

HerbertJordan:
This PR adds the ability to add tests including the restart of a network node.

Recently, issues have been reported that the restart of a node with local state fails during it startup phase. To address such issues, a new integration test including the restarting of nodes got added. One of the reported issue -- for the node failing to be able to re-load its local transaction pool data -- could be reproduced and fixed as part of this PR.


jenikd (approved):
Looks good, glad that it reporoduced the issue

LuisPH3 (approved):

PR #305, opened 2024-11-14
C/integration test tx order

cabrador:
This PR adds an integration test which checks the transaction order consistency within blocks.


tests/contracts/counter_event_emitter/counter_event_emitter.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.20;
+
+contract counter_event_emitter {
+    int private totalCount = 0;
+    mapping(address => int) public perAddrCount;
+    event Count(int count);
```
HerbertJordan, line 7: I think this should emit the `tot

HerbertJordan (approved):
Looks good, thanks.

PR #307, opened 2024-11-15
Integration of dynamic prices into price suggestion and TX pool

HerbertJordan:
This PR integrates base-fee based pricing into the transaction pool, event emitter, and the gas-price suggestion RPC. Among those, the following rules are enforced:
- for every block, a `baseFee` is computed based on the previous `baseFee` and the observed gas throughput (not part of this PR)
- gas price suggestions given through RPC calls are `lastBaseFee+10%`, where `lastBaseFee` is the fee of the most recent block
- the transaction pool accepts transactions with a minimum price of `lastBaseFee+5%`
- the event emitter is including transactions from the pool if they have a minimum price of `lastBaseFee+2%`
- the block processor computes the actual `baseFee` for the respective block and charges this price + tip as the `effectiveGasFee` to the user

The aim of this setup is to avoid pricing issues due to short-term base-fee variability. By design, the `baseFee` of a block should at most increase by ~1% per second.

### DIFF
--- a/eventcheck/epochcheck/epoch_check.go
+++ b/eventcheck/epochcheck/epoch_check.go
@@ -2,13 +2,11 @@ package epochcheck
 
 import (
 	"errors"
-	"math/big"
 
 	base "github.com/Fantom-foundation/lachesis-base/eventcheck/epochcheck"
 	"github.com/Fantom-foundation/lachesis-base/inter/idx"
 	"github.com/ethereum/go-ethereum/core/types"
 
-	"github.com/Fantom-foundation/go-opera/gossip/gasprice/gaspricelimits"
 	"github.com/Fantom-foundation/go-opera/inter"
 	"github.com/Fantom-foundation/go-opera/opera"
 )
@@ -83,11 +81,7 @@ func (v *Checker) checkGas(e inter.EventPayloadI, rules opera.Rules) error {
 	return nil
 }
 
-func CheckTxs(
-	txs types.Transactions,
-	rules opera.Rules,
-	baseFee *big.Int,
-) error {
+func CheckTxs(txs types.Transactions, rules opera.Rules) error {
 	maxType := uint8(0)
 	if rules.Upgrades.Berlin {
 		maxType = 1
@@ -98,20 +92,16 @@ func CheckTxs(
 	if rules.Upgrades.Sonic {
 		maxType = 3
 	}
-	limit := gaspricelimits.GetMinimumFeeCapForEventEmitter(baseFee)
 	for _, tx := range txs {
 		if tx.Type() > maxType {
 			return ErrUnsupportedTxType
-		}
-		if tx.GasFeeCapIntCmp(limit) < 0 {
-			return ErrUnderpriced
 		}
 	}
 	return nil
 }
 
-// Validate event. The baseFee may be nil if it should not be checked for.
-func (v *Checker) Validate(e inter.EventPayloadI, baseFee *big.Int) error {
+// Validate event
+func (v *Checker) Validate(e inter.EventPayloadI) error {
 	if err := v.Base.Validate(e); err != nil {
 		return err
 	}
@@ -129,7 +119,7 @@ func (v *Checker) Validate(e inter.EventPayloadI, baseFee *big.Int) error {
 	if err := v.checkGas(e, rules); err != nil {
 		return err
 	}
-	if err := CheckTxs(e.Txs(), rules, baseFee); err != nil {
+	if err := CheckTxs(e.Txs(), rules); err != nil {
 		return err
 	}
 
--- a/eventcheck/epochcheck/epoch_check.go
+++ b/eventcheck/epochcheck/epoch_check.go
@@ -110,7 +110,7 @@ func CheckTxs(
 	return nil
 }
 
-// Validate event; the baseFee may be nil if it should not be checked for.
+// Validate event. The baseFee may be nil if it should not be checked for.
 func (v *Checker) Validate(e inter.EventPayloadI, baseFee *big.Int) error {
 	if err := v.Base.Validate(e); err != nil {
 		return err
--- a/eventcheck/epochcheck/epoch_check.go
+++ b/eventcheck/epochcheck/epoch_check.go
@@ -2,11 +2,13 @@ package epochcheck
 
 import (
 	"errors"
+	"math/big"
 
 	base "github.com/Fantom-foundation/lachesis-base/eventcheck/epochcheck"
 	"github.com/Fantom-foundation/lachesis-base/inter/idx"
 	"github.com/ethereum/go-ethereum/core/types"
 
+	"github.com/Fantom-foundation/go-opera/gossip/gasprice/gaspricelimits"
 	"github.com/Fantom-foundation/go-opera/inter"
 	"github.com/Fantom-foundation/go-opera/opera"
 )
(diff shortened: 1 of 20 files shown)


LuisPH3:
> the event emitter is including transactions from the pool if they have a minimum price of `lastBaseFee+2%`

Is there is a mechanism to evict the transactions that fail this check? 
- A transaction is submitted with gas + 5%
- the transaction remains queued for 3 seconds (potential 3% price increase). Will not be allowed to proceed for completion, yet not possible to return an error to the client that submitted it because answered before. 
Transaction under these conditions need to be removed from queues (eventually? )  otherwise, the only way for a client to submit a new transaction into this validator is to pay the 10% "fine" of replacing the already queued one. 

On the other hand, not implementing the 2% constraint at this level removes transactions from the queue, (to be potentially ignored by the next level). 
It may be more robust, and easier to implement not to enforce this limit.

eventcheck/epochcheck/epoch_check.go
```diff
@@ -92,19 +98,20 @@ func CheckTxs(txs types.Transactions, rules opera.Rules) error {
 	if rules.Upgrades.Sonic {
 		maxType = 3
 	}
+	limit := gaspricelimits.GetMinimumFeeCapForEventEmitter(baseFee)
```
LuisPH3, line 101: There is a semantic change here, 
The previous code compared the price to a constant (which was already satisfied by the ingress validation) Therefore no tx would be prevented to be fetched by this reason. 
The new code may reject more transactions, and this could have unforeseen consequences of tx not being discarded. 
long story short: The previous code was redundant, the new one adds constraints.

evmcore/tx_pool.go
```diff
@@ -698,23 +698,22 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
...
 	// Ensure Opera-specific hard bounds
-	if recommendedGasTip, minPrice := pool.chain.EffectiveMinTip(), pool.chain.MinGasPrice(); recommendedGasTip != nil && minPrice != nil {
-		if tx.GasTipCap().Sign() < 0 {
-			log.Trace("Rejecting underpriced tx: negative tip", "GasTipCap", tx.GasTipCap())
-			return ErrUnderpriced
-		}
-		if tx.GasFeeCapIntCmp(new(big.Int).Add(recommendedGasTip, minPrice)) < 0 {
-			log.Trace("Rejecting underpriced tx: recommendedGasTip+minPrice", "recommendedGasTip", recommendedGasTip, "minPrice", minPrice, "tx.GasFeeCap", tx.GasFeeCap())
+	if baseFee := pool.chain.GetCurrentBaseFee(); baseFee != nil {
+		limit := gaspricelimits.GetMinimumFeeCapForTransactionPool(baseFee)
```
LuisPH3, line 710: this is way more elegant now.

LuisPH3 (approved):
Implementation seems correct to me, nevertheless I have my concerns about the pressure on the queue that the new introduced check for the tx that make into the next event. The reason being the previous check was always true, and transactions being rejected at this point for this reason was never tested before.

eventcheck/epochcheck/epoch_check.go
HerbertJordan: The previous was mostly redundant, yes. But `MaxGasPrice` is not a constant. It is a rule parameter that changes over time. This one in particular is altered on the Fantom mainnet after every epoch.

So also in the previous version, transactions that passed the incoming transaction-pool filter might get rejected at this point.

HerbertJordan:
> > the event emitter is including transactions from the pool if they have a minimum price of `lastBaseFee+2%`
> 
> Is there is a mechanism to evict the transactions that fail this check?
> 
> * A transaction is submitted with gas + 5%
> * the transaction remains queued for 3 seconds (potential 3% price increase). Will not be allowed to proceed for completion, yet not possible to return an error to the client that submitted it because answered before.
>   Transaction under these conditions need to be removed from queues (eventually? )  otherwise, the only way for a client to submit a new transaction into this validator is to pay the 10% "fine" of replacing the already queued one.
> 
> On the other hand, not implementing the 2% constraint at this level removes transactions from the queue, (to be potentially ignored by the next level). It may be more robust, and easier to implement, not to enforce this limit.

As discussed offline, we need to study the transaction pool's eviction policy as a follow-up task ([#76](https://github.com/Fantom-foundation/sonic-admin/issues/76)).

eventcheck/epochcheck/epoch_check.go
```diff
@@ -92,19 +98,20 @@ func CheckTxs(txs types.Transactions, rules opera.Rules) error {
...
+		if tx.GasFeeCapIntCmp(limit) < 0 {
 			return ErrUnderpriced
 		}
 	}
 	return nil
 }
 
-// Validate event
-func (v *Checker) Validate(e inter.EventPayloadI) error {
+// Validate event; the baseFee may be nil if it should not be checked for.
```
jenikd, line 113:
```suggestion
// Validate event. The baseFee may be nil if it should not be checked for.
```

tests/gas_price_suggestion_test.go
```diff
@@ -0,0 +1,178 @@
...
+	senderKey *ecdsa.PrivateKey
+	chainId   *big.Int
+}
+
+func (f *txFactory) makeLegacyTransactionWithPrice(
+	t *testing.T,
+	nonce uint64,
+	price int64,
+) *types.Transaction {
+	transaction, err := types.SignTx(types.NewTx(&types.AccessListTx{
```
jenikd, line 117: This creates AccessList transaction, not the legacy one

jenikd (changes requested):

thaarok (dismissed):
This keeps EthAPIBackend.SuggestGasTipCap() based on rules.Economy.MinGasPrice, which should be fixed yet, but we may keep this for a follow-up.
+the makeLegacyTransactionWithPrice test issue marked by Jenik

HerbertJordan:
@thaarok @jenikd I factored in your feedback, thanks for that. I also simplified the integration by removing the gas-price check from the event verification. In the end, it does not really fit there, and the risk of having nodes diverge because the event verification is inconsistent is not worth it.

Instead, the gas-fee limit check in the emitter is now using an already preexisting filter mechanism [here](https://github.com/Fantom-foundation/Sonic/blob/03b57450ee39d4d735844f3bdee1f0009d639e26/gossip/emitter/emitter.go#L270). This should have the same effect, but this part is processed locally, by each validator and does not affect the event validation.

gossip/ethapi_backend.go
```diff
@@ -481,7 +482,8 @@ func (b *EthAPIBackend) CurrentEpoch(ctx context.Context) idx.Epoch {
 }
 
 func (b *EthAPIBackend) MinGasPrice() *big.Int {
-	return b.state.MinGasPrice()
+	current := b.state.GetCurrentBaseFee()
+	return gaspricelimits.GetSuggestedGasPriceForNewTransactions(current)
 }
```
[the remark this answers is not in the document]
HerbertJordan: Should we update `SuggestGasTipCap()` to return a constant `0`?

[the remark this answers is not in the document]
thaarok: I think it should consider the current content of the txpool. Not sure.

thaarok (approved):
Not sure about the SuggestGasTipCap (which can be handled in a follow-up), otherwise OK.

gossip/ethapi_backend.go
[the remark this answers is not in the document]
HerbertJordan: Noted as a follow-up task ([#77](https://github.com/Fantom-foundation/sonic-admin/issues/77)).

jenikd (approved):

PR #312, opened 2024-11-15
Remove dependency on fjl/memsize library

jmpike:
As of Go 1.23, memsize no longer works because of a restriction added by the Go toolchain. The Go 1.23 compiler no longer allows access to runtime symbols via go:linkname, which prevents memsize from accessing the Stop-the-World functionality of the Go runtime.

The only reason for this library use was a Pprof endpoint providing an information about the amount of memory consumed by the node.Node structure. The usability of the endpoint is questionable and its removal seems to be the simplest way of restoring compatibility with the latest Go toolchain.


HerbertJordan (approved):
Looks good to me.

PR #316, opened 2024-11-18
Fix recording and reporting of epoch information in blocks

HerbertJordan:
Norma depends on the proper reporting of the epoch a block belongs for its logging and debugging. 

Before #301 this epoch information was extracted from a block's hash. Now, since this information is no longer encoded in the hash, an alternative source is required. This PR adds explicit tracking of epoch information.


jenikd (approved):
Looks good, thanks

PR #317, opened 2024-11-18
Introduce getAccount rpc method 

cabrador:
This PR adds `getAccount` method to the RPC according to [this](https://www.quicknode.com/docs/ethereum/eth_getAccount) documentation.

**Note:** As written in the code [here](https://github.com/Fantom-foundation/Sonic/blob/2001b3718c9182b159897adc0ceb66ba608dcb47/gossip/evmstore/carmen.go#L162) - Carmen currently actually does support `GetStorageRoot(addr)` - it returns a const non-empty hash for accounts which do not have empty storage and an empty hash for account which do have empty storage.

### DIFF
--- a/tests/get_account_test.go
+++ b/tests/get_account_test.go
@@ -1,0 +1,67 @@
+package tests
+
+import (
+	"github.com/Fantom-foundation/go-opera/ethapi"
+	"github.com/Fantom-foundation/go-opera/tests/contracts/transientstorage"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/holiman/uint256"
+	"testing"
+)
+
+func TestGetAccount(t *testing.T) {
+	net, err := StartIntegrationTestNet(t.TempDir())
+	if err != nil {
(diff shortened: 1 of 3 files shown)


tests/get_account_test.go
```diff
@@ -0,0 +1,67 @@
...
+	if res.StorageRoot == (common.Hash{}) {
+		t.Error("storage root should not be empty")
+	}
+
+	if got, want := (*uint256.Int)(res.Balance).Uint64(), uint64(0); got != want {
+		t.Errorf("balance too low, got: %d want: %d", got, want)
+	}
+
+	if res.Nonce < 1 {
+		t.Errorf("accounts nonce is expected to by at least 1")
```
jenikd, line 65:
```suggestion
		t.Errorf("account nonce is expected to by at least 1")
```

jenikd (approved):
Thanks

HerbertJordan (changes requested):
I agree with @jmpike, unless we provide a correct storage root, we should not offer this API endpoint.

cabrador:
I agree with you. Im closing this PR. We can re-open it once Carmen implements `GetStorageRoot` correctly.

PR #318, opened 2024-11-18
block/header properties tests

facuMH:
This PR adds test checking the following properties for `block` and `header`.
- check that the transaction root matches the transactions in the block
- check that the receipt root matches the receipts in the block
- check that the logs bloom matches the logs in the receipts
- coinbase is zero for all blocks
- difficulty and nonce is set to 0
- time is progressing strictly monotonically and approximately matches the current time
- the random `mixDigest` field is different for each block

NOTE: issues https://github.com/Fantom-foundation/sonic-admin/issues/81 and https://github.com/Fantom-foundation/sonic-admin/issues/80 were originally found during development of these tests.

### DIFF
--- a/tests/block_header_test.go
+++ b/tests/block_header_test.go
@@ -91,10 +91,16 @@ func TestBlockHeader_SatisfiesInvariants(t *testing.T) {
 		testHeaders_NonceIsZeroForAllBlocks(t, headers)
 	})
 
+	t.Run("TimeProgressesMonotonically", func(t *testing.T) {
+		testHeaders_TimeProgressesMonotonically(t, headers)
+	})
+
+	t.Run("MixDigestDiffersForAllBlocks", func(t *testing.T) {
+		testHeaders_MixDigestDiffersForAllBlocks(t, headers)
+	})
+
 	// TODO: Add more tests.
-	// - check that the receipt root matches the receipts in the block by hash
-	// - time is progressing strictly monotonically and approximately matches the current time
-	// - the random mixDigest field is different for each block
+	// - check that the receipt root matches the receipts in the block by hash (ISSUE #81)
 }
 
 func testHeaders_BlockNumberEqualsPositionInChain(t *testing.T, headers []*types.Header) {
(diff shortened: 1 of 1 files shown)


tests/block_header_test.go
```diff
@@ -155,3 +184,89 @@ func testHeaders_BaseFeeEvolutionFollowsPricingRules(t *testing.T, headers []*ty
...
+	for _, header := range headers {
+		require.Empty(header.Nonce.Uint64(), "nonce is not zero")
+	}
+}
+
+func testHeaders_TimeProgressesMonotonically(t *testing.T, headers []*types.Header) {
+	require := require.New(t)
+
+	for i := 1; i < len(headers); i++ {
+		require.GreaterOrEqual(headers[i].Time, headers[i-1].Time, "time is not monotonically increasing")
```
HerbertJordan, line 255: Nit: we do also store the nano-seconds in the block header, enabling us to get the block time at nano-second granularity (see [here](https://github.com/Fantom-foundation/Sonic/blob/2001b3718c9182b159897adc0ceb66ba608dcb47/tests/block_header_test.go#L113-L118)). Can you check that this time in each block is strictly greater than in its predecessor?

tests/block_header_test.go
```diff
@@ -155,3 +184,89 @@ func testHeaders_BaseFeeEvolutionFollowsPricingRules(t *testing.T, headers []*ty
...
+		logsBloom := types.CreateBloom(receipts)
+		require.Equal(header.Bloom, logsBloom, "logs bloom mismatch")
+	}
+}
+
+func testHeaders_CoinbaseIsZeroForAllBlocks(t *testing.T, headers []*types.Header) {
+	require := require.New(t)
+
+	for _, header := range headers {
+		require.Empty(header.Coinbase, "coinbase is not zero")
```
HerbertJordan, line 230: nit: Is `Empty` checking for a zero value here? It would be cleaner to do a `require.Equal` here.

HerbertJordan (approved):
Nice improvement and good catch. A few extra test cases and 2 additional bugs discovered!

LuisPH3 (approved):

tests/block_header_test.go
facuMH: yes, it does check for zero value. There is also a `.Zero` method, I'll use that one.

LuisPH3 (approved):

PR #321, opened 2024-11-19
Genesis block zero hash

facuMH:
This PR fixes https://github.com/Fantom-foundation/sonic-admin/issues/81. The missing hash is added by `s.SetBlockIndex` when initializing the genesis

### DIFF
--- a/evmcore/apply_fake_genesis.go
+++ b/evmcore/apply_fake_genesis.go
@@ -26,7 +26,7 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
-var FakeGenesisTime = inter.Timestamp( /*1608600000*/ time.Duration(time.Now().Unix()) * time.Second)
+var FakeGenesisTime = inter.Timestamp(time.Duration(time.Now().Unix()) * time.Second)
 
 // FakeKey gets n-th fake private key.
 func FakeKey(n uint32) *ecdsa.PrivateKey {
--- a/evmcore/apply_fake_genesis.go
+++ b/evmcore/apply_fake_genesis.go
@@ -26,7 +26,7 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
-var FakeGenesisTime = inter.Timestamp(1608600000 * time.Second)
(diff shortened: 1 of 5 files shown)


evmcore/apply_fake_genesis.go
```diff
@@ -26,7 +26,7 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
-var FakeGenesisTime = inter.Timestamp(1608600000 * time.Second)
+var FakeGenesisTime = inter.Timestamp(time.Duration(time.Now().Unix()) * time.Second)
```
HerbertJordan, line 29: I am not sure this will go well with Norma, where multiple nodes are initialized using the fake-genesis code independently and are then expected to cooperate with each other. If the time of the fake-genesis creation is used, as listed here, each of those nodes would already be out-of-sync from the start.

gossip/apply_genesis.go
```diff
@@ -60,14 +60,16 @@ func (s *Store) ApplyGenesis(g genesis.Genesis) (err error) {
 	rules := s.GetRules()
 	gasLimit := rules.Blocks.MaxBlockGas
 
-	s.SetBlock(0, inter.NewBlockBuilder().
+	blockZero := inter.NewBlockBuilder().
 		WithNumber(0).
-		WithTime(evmcore.FakeGenesisTime-1). // TODO: extend genesis generator to provide time
+		WithTime(evmcore.FakeGenesisTime - 1).
```
HerbertJordan, line 65: It would be better to add some way to extract this information from `genesis.Genesis` than to use a non-constant value.

HerbertJordan (changes requested):
The block-hash fix looks good, the genesis-block timestamp however, likely leads to problems.

evmcore/apply_fake_genesis.go
facuMH: ok. I'll remove the timestamp changes from this PR so that we can merge thee hash fix and we keep discussing a better way to approach the timestamp issue.

HerbertJordan (approved):

PR #323, opened 2024-11-19
Switch to fixed emitter rate

HerbertJordan:
This PR introduces a simplification of the event emitter timing policy. The previous emitter policy which was deriving the decision whether to emit or not from various network-state parameters is (optionally) replaced by a simpler time-based emitter rate.

#### Fixed-Rate Emitter Policy
If enabled, the new emitter has the following properties:
- in normal mode, every validator produces a new event every `X` milliseconds
- if for threshold time `T` no new events are confirmed on the network, the emitter switches into a `stall` mode and reduces the emitter rate to `Y`

The parameters `X`, `T`, and `Y` can be configured using network rules. Those rules also allow to disable this rate-based emitter policy and revert to the previous policy.

#### Evaluation
In Norma based experiments, this new emitter timing policy helps reducing block-time variability. To demonstrate this, the following scenario was evaluated:
- a network is started comprising 2 validators
- for 40 seconds a minimal load of 1 transaction per second is send into the network, produced by a single user
- after a 10 second break, a load of 10 transactions per second is send into the network for another 40 seconds

The following chart shows the block-times **before** this change:
image

For the low-load interval, few blocks are created, with a block time of approximately 4.6 seconds. Once the 10 Tx/s load case is reached, numerous blocks are created, with a block time of a few 100 milliseconds -- significantly lower than required.

Thus, the network load had a considerable impact on the block time -- and thus the lower boundary of the time to finality. With low load, blocks could take more than 4 seconds to be created, while with moderate load too many blocks are created.

The following chart shows the block-times **after** this change, with a target-block time of 600 ms (dashed red line):
image

With this change, the timing of new blocks becomes independent of the actual network traffic.

When repeating the experiment with 20 validator nodes, block times are also found to be stable, as shown in the following graph:
image

However, due to the delay of the propagation of events and the limited degree of parent references in events, the block time becomes more variable. For the network deployment, this property should be monitored and the desired emitter rate adjusted accordingly.

### DIFF
--- a/gossip/c_block_callbacks.go
+++ b/gossip/c_block_callbacks.go
@@ -158,6 +158,11 @@ func consensusCallbackBeginBlockFn(
 				// We have to skip block in case (1) to ensure that every block ID is unique.
 				// If Atropos ID wasn't used as a block ID, it wouldn't be required.
 				skipBlock := atroposDegenerate
+				// Check if empty block should be pruned
+				if es.Rules.Emitter.Interval == 0 {
+					emptyBlock := false // confirmedEvents.Len() == 0 && cBlock.Cheaters.Len() == 0
+					skipBlock = skipBlock || (emptyBlock && blockCtx.Time < bs.LastBlock.Time+es.Rules.Blocks.MaxEmptyBlockSkipPeriod)
+				}
 				// Finalize the progress of eventProcessor
 				bs = eventProcessor.Finalize(blockCtx, skipBlock) // TODO: refactor to not mutate the bs, it is unclear
 				if skipBlock {
--- a/gossip/c_block_callbacks.go
+++ b/gossip/c_block_callbacks.go
@@ -158,9 +158,6 @@ func consensusCallbackBeginBlockFn(
 				// We have to skip block in case (1) to ensure that every block ID is unique.
 				// If Atropos ID wasn't used as a block ID, it wouldn't be required.
 				skipBlock := atroposDegenerate
-				// Check if empty block should be pruned
-				emptyBlock := confirmedEvents.Len() == 0 && cBlock.Cheaters.Len() == 0
-				skipBlock = skipBlock || (emptyBlock && blockCtx.Time < bs.LastBlock.Time+es.Rules.Blocks.MaxEmptyBlockSkipPeriod)
 				// Finalize the progress of eventProcessor
 				bs = eventProcessor.Finalize(blockCtx, skipBlock) // TODO: refactor to not mutate the bs, it is unclear
 				if skipBlock {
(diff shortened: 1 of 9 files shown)


Tjofil:
Simplification of emitting rule is welcome and code changes look good. A mild concern:
Will block skipping logic 'undo' some of the emission changes made here ? In low tx scenarios, events with no txs, now being regularly emitted, would dominate and cause blocks to be skipped. Have you rerun the simulations after you've brought back the block skipping mechanism, I'm curious to see how it would play out.

Generally, I feel a bit unconvinced about influencing the block emission rate through event emission rate. A lot of intermediate logic (which also may evolve drastically in the future) between these two can make the event emission rate irrelevant or maybe backfire? One example is the block skipping logic, other being Consensus with strong asynchronous behavior. Going towards synchronous-ish requirements may close the opportunities for more radical TTF gains on the asynchronous side. What's your opinion on this ?

HerbertJordan:
@Tjofil 
> Will block skipping logic 'undo' some of the emission changes made here ? In low tx scenarios, events with no txs, now being regularly emitted, would dominate and cause blocks to be skipped. Have you rerun the simulations after you've brought back the block skipping mechanism, I'm curious to see how it would play out.

I re-run the evaluation with and without the empty-block filter. Here are the results with the empty-block filter:
image

And here without (current state of the PR):
image

The latter shows a curios property of our current system: right now, the transaction pool is set up to only ever have a single transaction per user within an unconfirmed event. Only once the event including the transaction is confirmed, another one is admitted in a new event. This leads to the effect that for a single user sending in 1 Tx/s some burstiness is introduced, frequently leading to 2 Tx ending up in the same block, as seen here:
image

If this 1 Tx/s load is distributed among multiple users, this effect is not active, leading to a smooth load distribution of 1 Tx/s:
image
and also reduced block times, since less empty blocks are filtered out in-between:
image

HerbertJordan:
@Tjofil 
> Edit: What's the ratio between number of emitted events with emission policy and without ? What's the percentage of hollow events in each?

We do not have explicit metrics for that, but I could extract the following information from the logs:

| Version |  Events Emitted | Empty Events Emitted |
|:---:|:---:|:---:|
| base |  530 | 197 |
| mod |  470 |  296 |

To be considered here is: the majority of events in the `base` case are produced in the later 40 seconds of the experiment while with the modifications of this PR, the number of events is evenly distributed over the full runtime. This also makes the event generation rate predictable.

Lots of heuristics this PR is eliminating is aiming for minimizing the load on a low-utilized network where only very few transactions are processed. I would argue that we should not optimize for this use case.

HerbertJordan:
@Tjofil 

> Generally, I feel a bit unconvinced about influencing the block emission rate through event emission rate. A lot of intermediate logic (which also may evolve drastically in the future) between these two can make the event emission rate irrelevant or maybe backfire? One example is the block skipping logic, other being Consensus with strong asynchronous behavior. Going towards synchronous-ish requirements may close the opportunities for more radical TTF gains on the asynchronous side. What's your opinion on this ?

I would like to arrange a meeting discussing these points with you.

PR #327, opened 2024-11-22
Lower integration tests memory usage

cabrador:
This PR lowers memory usage for fake net and integration tests by ~500 MB per test.
Before:
Screenshot 2024-11-28 at 12 36 47

After:
Screenshot 2024-11-28 at 12 37 05

part of https://github.com/Fantom-foundation/sonic-admin/issues/50

### DIFF
--- a/cmd/sonicd/app/launcher.go
+++ b/cmd/sonicd/app/launcher.go
@@ -65,6 +65,8 @@ func initFlags() {
 	}
 	performanceFlags = []cli.Flag{
 		flags.CacheFlag,
+		flags.LiveDbCacheFlag,
+		flags.ArchiveCacheFlag,
 	}
 	networkingFlags = []cli.Flag{
 		flags.BootnodesFlag,
@@ -222,9 +224,12 @@ func lachesisMain(ctx *cli.Context) error {
 	}
 
 	metrics.SetDataDir(cfg.Node.DataDir) // report disk space usage into metrics
-	if ctx.GlobalIsSet(config.FakeNetFlag.Name) {
-		cfg.OperaStore.EVM.StateDb.LiveCache = 1
-		cfg.OperaStore.EVM.StateDb.ArchiveCache = 1
+	if ctx.IsSet(flags.LiveDbCacheFlag.Name) {
+		cfg.OperaStore.EVM.StateDb.LiveCache = ctx.Int64(flags.LiveDbCacheFlag.Name)
+	}
+
+	if ctx.IsSet(flags.ArchiveCacheFlag.Name) {
+		cfg.OperaStore.EVM.StateDb.ArchiveCache = ctx.Int64(flags.ArchiveCacheFlag.Name)
 	}
 
 	node, _, nodeClose, err := config.MakeNode(ctx, cfg)
--- a/cmd/sonicd/app/launcher.go
+++ b/cmd/sonicd/app/launcher.go
@@ -222,6 +222,10 @@ func lachesisMain(ctx *cli.Context) error {
 	}
 
 	metrics.SetDataDir(cfg.Node.DataDir) // report disk space usage into metrics
+	if ctx.GlobalIsSet(config.FakeNetFlag.Name) {
+		cfg.OperaStore.EVM.StateDb.LiveCache = 1
+		cfg.OperaStore.EVM.StateDb.ArchiveCache = 1
+	}
 
 	node, _, nodeClose, err := config.MakeNode(ctx, cfg)
 	if err != nil {
--- a/cmd/sonicd/app/launcher.go
+++ b/cmd/sonicd/app/launcher.go
@@ -224,12 +224,14 @@ func lachesisMain(ctx *cli.Context) error {
 	}
 
 	metrics.SetDataDir(cfg.Node.DataDir) // report disk space usage into metrics
-	if ctx.IsSet(flags.LiveDbCacheFlag.Name) {
-		cfg.OperaStore.EVM.StateDb.LiveCache = ctx.Int64(flags.LiveDbCacheFlag.Name)
+	liveCache := ctx.Int64(flags.LiveDbCacheFlag.Name)
+	if liveCache > 0 {
+		cfg.OperaStore.EVM.StateDb.LiveCache = liveCache
 	}
 
-	if ctx.IsSet(flags.ArchiveCacheFlag.Name) {
-		cfg.OperaStore.EVM.StateDb.ArchiveCache = ctx.Int64(flags.ArchiveCacheFlag.Name)
+	archiveCache := ctx.Int64(flags.ArchiveCacheFlag.Name)
+	if archiveCache > 0 {
+		cfg.OperaStore.EVM.StateDb.ArchiveCache = archiveCache
 	}
 
 	node, _, nodeClose, err := config.MakeNode(ctx, cfg)
--- a/cmd/sonicd/app/launcher.go
+++ b/cmd/sonicd/app/launcher.go
@@ -224,12 +224,12 @@ func lachesisMain(ctx *cli.Context) error {
 	}
 
 	metrics.SetDataDir(cfg.Node.DataDir) // report disk space usage into metrics
-	liveCache := ctx.Int64(flags.LiveDbCacheFlag.Name)
+	liveCache := ctx.GlobalInt64(flags.LiveDbCacheFlag.Name)
 	if liveCache > 0 {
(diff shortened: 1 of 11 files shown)


LuisPH3 (dismissed):

cmd/sonictool/db/dbutils.go
```diff
@@ -61,7 +61,7 @@ func MakeDbProducer(chaindataDir string, cacheRatio cachescale.Func) (kvdb.FullD
 	})
 }
 
-func MakeGossipDb(dbs kvdb.FullDBProducer, dataDir string, validatorMode bool, cacheRatio cachescale.Func) (*gossip.Store, error) {
+func MakeGossipDb(dbs kvdb.FullDBProducer, dataDir string, validatorMode bool, cacheRatio cachescale.Func, isFakeNet bool) (*gossip.Store, error) {
```
HerbertJordan, line 64: nit: try to avoid boolean flags. They do harm readability on the caller's end significantly since they do not provide any hint what `true` or `false` is supposed to mean.

cmd/sonicd/app/launcher.go
```diff
@@ -222,6 +222,10 @@ func lachesisMain(ctx *cli.Context) error {
 	}
 
 	metrics.SetDataDir(cfg.Node.DataDir) // report disk space usage into metrics
+	if ctx.GlobalIsSet(config.FakeNetFlag.Name) {
```
HerbertJordan, line 225: As in #334, I do not think it is a good idea to link the cache configuration with the fake-net flag. This would prohibit scalability tests on fake-nets.

HerbertJordan (changes requested):

config/flags/flags.go
```diff
@@ -348,4 +348,14 @@ var (
 		Name:  "lachesis.suppress-frame-panic",
 		Usage: "Suppress frame missmatch error (when testing on historical imported/synced events)",
 	}
+
+	// StateDb
+	LiveDbCacheFlag = cli.Int64Flag{
+		Name:  "statedb.livecache",
+		Usage: "Size of live db cache in bytes.",
```
LuisPH3, line 355: Please document: 
- what is the default value?  This is needed, if not default, then a recommended value should be provided.
- does it accept 0 or -1 with special semantics? 

Users need to understand if they need to modify this values or if it is safe to ignore them.

config/flags/flags.go
```diff
@@ -348,4 +348,14 @@ var (
...
+	}
+	ArchiveCacheFlag = cli.IntFlag{
+		Name:  "statedb.archivecache",
+		Usage: "Size of archive cache in bytes.",
```
LuisPH3, line 359: Please document:
- what is the default value? 
- does it accept 0 or -1 with special semantics?

tests/integration_test_net.go
```diff
@@ -65,7 +65,15 @@ func StartIntegrationTestNet(directory string) (*IntegrationTestNet, error) {
 	// initialize the data directory for the single node on the test network
 	// equivalent to running `sonictool --datadir <dataDir> genesis fake 1`
 	originalArgs := os.Args
-	os.Args = []string{"sonictool", "--datadir", result.stateDir(), "genesis", "fake", "1"}
+	os.Args = []string{
+		"sonictool",
+		"--datadir", result.stateDir(),
+		"genesis",
+		"fake",
+		"--statedb.livecache", "1",
```
LuisPH3, line 73: 1 byte? why not 0 ?

tests/integration_test_net.go
```diff
@@ -65,7 +65,15 @@ func StartIntegrationTestNet(directory string) (*IntegrationTestNet, error) {
...
+		"--statedb.archivecache", "1",
```
LuisPH3, line 74: same here, 1 byte? is the scale correct?

cmd/sonictool/db/dbutils.go
```diff
@@ -61,15 +63,23 @@ func MakeDbProducer(chaindataDir string, cacheRatio cachescale.Func) (kvdb.FullD
 	})
 }
 
-func MakeGossipDb(dbs kvdb.FullDBProducer, dataDir string, validatorMode bool, cacheRatio cachescale.Func) (*gossip.Store, error) {
+func MakeGossipDb(ctx *cli.Context, dbs kvdb.FullDBProducer, dataDir string, validatorMode bool, cacheRatio cachescale.Func) (*gossip.Store, error) {
```
LuisPH3, line 66: This is odd, but I see that it is some kind of pattern in this part of the code:
Passing around the cli.context is like sharing the command line arguments over. Code consuming some flag will retrieve the value from the command line arguments and validate it in situ. This may be too late to provide meaningful error reports to the user to correct a faulty command line entry.

cmd/sonictool/db/dbutils.go
```diff
@@ -31,7 +33,7 @@ func makeDatabaseHandles() uint64 {
 	if err != nil {
 		panic(fmt.Errorf("failed to raise file descriptor allowance: %v", err))
 	}
-	return raised / 6 + 1
+	return raised/6 + 1
```
LuisPH3, line 36: ? your ide formats it different?

LuisPH3 (changes requested):
Unfortunately, the mechanism used to forward the cache settings into the algorithms that need it is not ok.

`func MakeGossipDb(ctx *cli.Context, dbs kvdb.FullDBProducer, dataDir string, validatorMode bool, cacheRatio cachescale.Func) (*gossip.Store, error) {`

This function already includes parameters deduced from the command line, additionally this is too late to parse command line arguments, ranges are not checked, and faulty values will be propagated to internal components for undefined behavior. 


Please perform the command line arguments parsing the outermost possible layer of the program, sanitize arguments and return an error to the user as early as possible. 

To avoid aliasing arguments into the program logic functions (like `func x( bool, int , float ,int`) I propose alternatives: 

1. pack all configuration arguments from a function into a struct (this way you can add names to the funciton call side)
```
callFunction(configuration{
     argument1: 1,
     argument2: 2,
     })
```
2. Pack single arguments into a struct (same effect, naming one argument at the time).

3. Use a builder pattern (may be a little too heavy for this use case).

tests/integration_test_net.go
cabrador: There is not any reason in particular, I just saw we use this value already in tests. Both 0 and 1 sets lowest value in Carmen which is 2 KB

tests/integration_test_net.go
cabrador: There is not any reason in particular, I just saw we use this value already in tests. Both 0 and 1 sets lowest value in Carmen which is 2 KB

cmd/sonictool/db/dbutils.go
cabrador: Yes

cabrador: Removed.

LuisPH3 (approved):

cmd/sonictool/chain/export_events.go
```diff
@@ -24,15 +26,22 @@ var (
 // always print out progress. This avoids the user wondering what's going on.
 const statsReportLimit = 8 * time.Second
 
-func ExportEvents(w io.Writer, dataDir string, from, to idx.Epoch) (err error) {
+func ExportEvents(ctx *cli.Context, w io.Writer, dataDir string, from, to idx.Epoch) (err error) {
```
HerbertJordan, line 29: nit: passing the full context to the `ExportEvents` and `ImportGenesisStore` is maybe over-sized and it moves the interpretation of flags into the implementations of those functions. I would instead create a storage-config struct including the cache sizes and maybe the data-dir. Wdyt?

HerbertJordan (changes requested):

cmd/sonictool/chain/export_events.go
cabrador: I dont like adding another struct for config - there is quite a mess in that. How about I use the already created struct `GossipDbParameters` and just pass in the Cache sizes and dir and fill the rest  of the struct later?

LuisPH3: generating `GossipDbParameters` from the command line parsing could be an option. But if this turns cumbersome, please use a new struct with two fields for
```
	LiveDbCache:   ctx.Int64(flags.LiveDbCacheFlag.Name),
	ArchiveCache:  ctx.Int64(flags.ArchiveCacheFlag.Name),
```

HerbertJordan (dismissed):

HerbertJordan (approved):

LuisPH3:
:partying_face:

PR #328, opened 2024-11-22
Remove transaction pool gas price limit configuration flag and logic.

LuisPH3:
This PR removes the minimum gas price configuration parameter from the transaction pool.

The purpose of this parameter is to define the minimum fee a pool requires to accept a transaction. Since the introduction of the pool new pricing [schema](https://github.com/Fantom-foundation/Sonic/commit/84b9036bfc352c8d3c89a465799df7414bfc7694), this parameter is obsoleted. 

The existing implementation would mix gas price (Absolute measure)  and gas tips (an extra added on top of the base fee). The changes in this pr enable the following properties: 

- Transaction pool validation shall allow zero tip transactions: 
    - Validators in the sonic network do not cash the tip and define no minimum tipping.
- Tipping provides priority for the transaction: 
    - Remote transactions will not be dropped because of insufficient tipping, as there is no lower limit.
- Transaction pool shall not evict transactions because of insufficient tipping. 
    - Same reason as the point before

Partially enables: 
- Transactions shall not be default to tip, and tipping shall not be recommended 
    - Te oracle has still some heuristics that may need a new revision


evmcore/tx_pool.go
```diff
@@ -445,28 +437,9 @@ func (pool *TxPool) GasPrice() *big.Int {
...
-		// pool.priced is sorted by GasFeeCap, so we have to iterate through pool.all instead
-		drop := pool.all.RemotesBelowTip(price)
-		for _, tx := range drop {
-			pool.removeTx(tx.Hash(), true)
-		}
-		pool.priced.Removed(len(drop))
-	}
-
-	log.Info("Transaction pool price threshold updated", "price", price)
+	return new(big.Int).Set(
```
HerbertJordan, line 440: The `GasPrice` function seems to be nowhere used any more. The only reference is in the `gossip.TxPool` interface, but there it can be deleted as well as it is not used anywhere.

evmcore/tx_pool_test.go
```diff
@@ -1541,129 +1548,130 @@ func TestTransactionPendingMinimumAllowance(t *testing.T) {
 	}
 }
 
+// TODO: this test is rewritten in replacements test
```
HerbertJordan, line 1551: Left-over TODO / missing code cleanup?

Also: where is this replacement test?

HerbertJordan (changes requested):
I think it looks good, but I would like to also get the approval by @jmpike for this kind of changes.

evmcore/tx_pool_test.go
LuisPH3: here : https://github.com/Fantom-foundation/Sonic/pull/329

evmcore/tx_pool.go
LuisPH3: It is required by the interface to the gossip protocol:  https://github.com/Fantom-foundation/Sonic/blob/63190cb72ba47ca67ce8162046ab0b0cc18a76de/gossip/protocol.go#L113

LuisPH3:
this change is not required.

PR #329, opened 2024-11-22
Remove transaction pool price bump parameter

LuisPH3:
This PR removes the price bump parameter required to replace transactions in the pool.

The transaction pool performs a check before a transaction can be replaced. This happens when a transaction has been in the transaction pool for some time, it hasn’t been assigned to any event, and the user wants to reconfigure it (submit a new transaction with the same Sender account and Nonce).
The current mechanism is insufficient because of the following reasons:
- Although one of the two maximum gas fee or maximum priority fee have to be increased by a configured percentage, first if in the function forces the second parameter to be increased as well, which is counter-intuitive. 
- Configured percentage increment is applied to the previous submitted values, because the maximum priority fee could be zero, a zero increment would be accepted.
- Incrementing maximum gas fee does not provide any advantage: the transaction has been already validated, and therefore it includes a maximum gas fee of 105% of the current base fee. If the policy pursues to penalize replacements, increasing this parameter will render no penalty, since effective gas price may be lower than maximum priority fee, and therefore it may be lower than the previous transaction value for this parameter.

Transaction replacement needs to remain antisymmetric, to define an order over the replaced transactions. Replacements are therefore deterministic and independent of the order in which the transactions received for any pool of the network.

This change replaces the current algorithm with a version where the transaction is accepted if the maximum priority fee is improved by a configurable increment. Special handling is required when the old transaction defined zero maximum priority fee, in which case a new maximum priority fee larger than the last value should suffice.


evmcore/tx_list.go
```diff
@@ -282,33 +282,36 @@ func (l *txList) Overlaps(tx *types.Transaction) bool {
...
+	// tips to stablish an order. This order makes replacements deterministic in
+	// each replacement in the distributed pools.
 	if old != nil {
-		if old.GasFeeCapCmp(tx) >= 0 || old.GasTipCapCmp(tx) >= 0 {
+		oldTipCap := old.GasTipCap()
+		newTipCap := tx.GasTipCap()
+
+		numerator := big.NewInt(100 + int64(priceBump))
+		denominator := big.NewInt(100)
+		minimumIncrement := new(big.Int).Div(new(big.Int).Mul(oldTipCap, numerator), denominator)
```
HerbertJordan, line 299: nit: `minimumIncrement` => `minimumNewTipCap`

evmcore/tx_list.go
```diff
@@ -282,33 +282,36 @@ func (l *txList) Overlaps(tx *types.Transaction) bool {
...
-		thresholdFeeCap := aFeeCap.Div(aFeeCap, b)
-		thresholdTip := aTip.Div(aTip, b)
-
-		// Have to ensure that either the new fee cap or tip is higher than the
-		// old ones as well as checking the percentage threshold to ensure that
-		// this is accurate for low (Wei-level) gas price replacements
-		if tx.GasFeeCapIntCmp(thresholdFeeCap) < 0 || tx.GasTipCapIntCmp(thresholdTip) < 0 {
+
+		// if the new transaction has a lower tip than the old one, it can't replace it
+		// unless the new tip is at least x% higher than the old one
```
HerbertJordan, line 309:
```suggestion
		// if the new transaction has an insufficient tip cap, it cannot replace the old one
```

evmcore/tx_list.go
```diff
@@ -282,33 +282,36 @@ func (l *txList) Overlaps(tx *types.Transaction) bool {
...
+		if newTipCap.Cmp(minimumIncrement) < 0 {
```
HerbertJordan, line 310: While reviewing this, I was thinking about the following possible attack vector for flooding the network with replacement requests:
- i send in a transaction with `MaxFeeCap` = 100 and `MaxTipCap`=1
- i now keep sending in transactions where `MaxFeeCap` = 100 and the `MaxTipCap` grows by 10% each time; I can send in hundreds of those, growing the tip-cap into astronomical ranges, without risking any high prices, since the `MaxFeeCap` remains constant;

Is a transaction with `MaxTipCap` > `MaxFeeCap` valid? If not, I think we need to demand a X% increase in the `MaxFeeCap` as well. Otherwise, the "promised" Tip would never materialize.

evmcore/tx_list_test.go
```diff
@@ -38,7 +39,7 @@ func TestStrictTxListAdd(t *testing.T) {
 	// Insert the transactions in a random order
 	list := newTxList(true)
 	for _, v := range rand.Perm(len(txs)) {
-		list.Add(txs[v], DefaultTxPoolConfig.PriceBump)
+		list.Add(txs[v], 10)
```
HerbertJordan, line 42: nit: this replacement seems unnecessary. The `PriceBump` flag was not removed.

evmcore/tx_list_test.go
```diff
@@ -64,7 +65,34 @@ func BenchmarkTxListAdd(t *testing.B) {
 	priceLimit := big.NewInt(int64(DefaultTxPoolConfig.PriceLimit))
 	t.ResetTimer()
 	for _, v := range rand.Perm(len(txs)) {
-		list.Add(txs[v], DefaultTxPoolConfig.PriceBump)
-		list.Filter(priceLimit, DefaultTxPoolConfig.PriceBump)
+		list.Add(txs[v], 10)
```
HerbertJordan, line 68: same here

evmcore/tx_pool_test.go
```diff
@@ -2350,115 +2354,89 @@ func TestTransactionReplacement(t *testing.T) {
...
-		}
+	tests := map[string]struct {
+		originalTx    *types.Transaction
+		replacementTx *types.Transaction
+		expectedErr   error
+	}{
+		"Reject not bumping tip and fee cap": {
+			originalTx:    dynamicFeeTx(0, 100000, big.NewInt(gasFeeCap), big.NewInt(gasTipCap), key),
+			replacementTx: dynamicFeeTx(0, 100001, big.NewInt(gasFeeCap), big.NewInt(gasTipCap), key),
+			expectedErr:   ErrReplaceUnderpriced, // current lists accept any tip increment increment
```
HerbertJordan, line 2376: nit: I do not understand the comment in this line ...

evmcore/tx_pool_test.go
```diff
@@ -227,6 +228,9 @@ func validateTxPoolInternals(pool *TxPool) error {
 func validateEvents(events chan NewTxsNotify, count int) error {
 	var received []*types.Transaction
 
+	// add a timer to detect non-firing events
+	time.Sleep(50 * time.Millisecond)
```
HerbertJordan, line 232: not sure I understand this ... the following loop seems to block on incoming events and has a 5-second timeout. This sleep step seems to be unnecessary. Maybe the time limit in line 249 should be raised instead?

evmcore/tx_pool_test.go
```diff
@@ -2350,115 +2354,89 @@ func TestTransactionReplacement(t *testing.T) {
...
+			err := pool.AddRemote(test.originalTx)
+			require.NoError(t, err)
+
+			err = pool.AddRemote(test.replacementTx)
+			require.Equal(t, test.expectedErr, err)
+		})
 	}
 }
 
+func TestTransactionPool_FeedAnnouncesChanges(t *testing.T) {
```
HerbertJordan, line 2406: Nice cleanup of those two test cases, thanks!

HerbertJordan (changes requested):

evmcore/tx_list.go
LuisPH3: Right. This was my first thought, and therefore simplified everything because no policy here was satisfactory. 

`MaxTipCap > MaxFeeCap` is an error: https://github.com/Fantom-foundation/Sonic/blob/63190cb72ba47ca67ce8162046ab0b0cc18a76de/evmcore/tx_pool.go#L694

Because `maxTipCap` cannot be larger than the `maxFeeCap`, transactions that reach this limit need to rise are rejected before replacement during validation. The user can re-submit the transaction with a larger `maxTipCap` to increase the `maxFeeCap`, this check is implicit.

evmcore/tx_pool_test.go
LuisPH3: The problem here is that without a sleep, the check will happen right away after calling the function.
If there is a background routine that should not trigger an event, but it does. The test will validate the event before the faulty behavior happens. This complete test is a little weak in this regard. But I think that it does not worth not to rewrite it now.

PR #333, opened 2024-11-24
Extend genesis block record with missing data

jmpike:
The fixed block hash is not compatible with the reduced `LLRFullBlockRecord` used within the genesis snapshot. If the block database is reconstructed from a genesis file, it does not contain the same chain of blocks due to the missing attributes. This leads to local network split and failed synchronization after the next epoch is reached.

This PR introduces an extended version of the block record along with the use of the Block builder to reconstruct the exactly same state from the Genesis as if the block was created from the chain progression itself. The expected block hash is stored along with the LLR record and used to verify that the reconstructed block matches the exported one.

The PR changes genesis transaction execution to correctly build the full block record. This leads to a different genesis block state record and so makes the new genesis processing incompatible with the previous versions.

### DIFF
--- a/cmd/sonicd/app/misccmd.go
+++ b/cmd/sonicd/app/misccmd.go
@@ -2,9 +2,10 @@ package app
 
 import (
 	"fmt"
-	"github.com/Fantom-foundation/go-opera/version"
 	"os"
 	"runtime"
+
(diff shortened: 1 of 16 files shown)


thaarok (approved):

HerbertJordan (approved):
Looks good to me.

PR #334, opened 2024-11-25
Lower memory consumption for integration tests.

cabrador:
This PR lowers memory consumption by fakenet and integration tests by StateDB instances by ~250 MB
Before
Screenshot 2024-11-25 at 9 36 00

After:
Screenshot 2024-11-25 at 9 36 41

part of https://github.com/Fantom-foundation/sonic-admin/issues/50

### DIFF
--- a/cmd/sonicd/app/launcher.go
+++ b/cmd/sonicd/app/launcher.go
@@ -66,6 +66,7 @@ func initFlags() {
 		flags.CacheFlag,
 		flags.LiveDbCacheFlag,
 		flags.ArchiveCacheFlag,
+		flags.StateDbCacheCapacityFlag,
 	}
 	networkingFlags = []cli.Flag{
 		flags.BootnodesFlag,
--- a/config/flags/flags.go
+++ b/config/flags/flags.go
@@ -367,8 +367,9 @@ var (
 	}
 	StateDbCacheCapacityFlag = cli.IntFlag{
 		Name: "statedb.cache",
-		Usage: "Size of StateDb instances cache in bytes. Leaving this blank (which is generally recommended), or setting" +
-			"this to <1 will automatically set the cache capacity a hard-coded constant.",
+		Usage: "The number of cached data elements by each StateDb instance. Since this is an experimental feature " +
+			"used mainly by tests it is generally recommended leaving this blank. In tests no value lower than 1024 " +
+			"is recommended. Setting this to <1 will automatically set the cache capacity to a DB defined default value.",
 		Value: 0,
 	}
 )
--- a/config/flags/flags.go
+++ b/config/flags/flags.go
@@ -365,4 +365,10 @@ var (
 			"Setting this value to <=2000 will result in pre-confugired cache capacity of 2KB", CacheFlag.Name),
 		Value: 0,
 	}
+	StateDbCacheCapacityFlag = cli.IntFlag{
+		Name: "statedb.cache",
+		Usage: "Size of StateDb instances cache in bytes. Leaving this blank (which is generally recommended), or setting" +
+			"this to <1 will automatically set the cache capacity a hard-coded constant.",
+		Value: 0,
+	}
 )
--- a/integration/makegenesis/genesis.go
+++ b/integration/makegenesis/genesis.go
@@ -113,7 +113,7 @@ func (b *GenesisBuilder) CurrentHash() hash.Hash {
 	return er.Hash()
 }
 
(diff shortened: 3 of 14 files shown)


cmd/sonicd/app/launcher.go
```diff
@@ -225,6 +225,7 @@ func lachesisMain(ctx *cli.Context) error {
 	if ctx.GlobalIsSet(config.FakeNetFlag.Name) {
 		cfg.OperaStore.EVM.StateDb.LiveCache = 1
 		cfg.OperaStore.EVM.StateDb.ArchiveCache = 1
+		cfg.OperaStore.EVM.Cache.StateDbCapacity = 1024
```
HerbertJordan, line 228: I would favor an explicit flag for the state-db capacity that we set in our integration tests instead of limiting the cache size in general for fake-nets.

The reason is: Norma or other testing tools may employ fake-net setups also for scalability and/or performance tests. In those cases, the given capacity may be too small.

HerbertJordan (changes requested):
I like the memory reduction, but I think the StateDB capacity should be explicitly controlled by some flag, not implicitly through the fake-net switch.

config/flags/flags.go
```diff
@@ -348,4 +348,9 @@ var (
 		Name:  "lachesis.suppress-frame-panic",
 		Usage: "Suppress frame missmatch error (when testing on historical imported/synced events)",
 	}
+
+	StateDbCacheCapacitySize = cli.IntFlag{
+		Name:  "statedb.cache",
+		Usage: "Cache size for StateDb instances",
```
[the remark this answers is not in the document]
cabrador: Done. Hope the words I used are well formulated.

LuisPH3 (dismissed):

config/flags/flags.go
```diff
@@ -365,4 +365,10 @@ var (
 			"Setting this value to <=2000 will result in pre-confugired cache capacity of 2KB", CacheFlag.Name),
 		Value: 0,
 	}
+	StateDbCacheCapacityFlag = cli.IntFlag{
+		Name: "statedb.cache",
+		Usage: "Size of StateDb instances cache in bytes. Leaving this blank (which is generally recommended), or setting" +
```
HerbertJordan, line 370: This cache parameter is actually not in bytes but in number of elements in the storage data cache created [here](https://github.com/Fantom-foundation/Carmen/blob/690bc10fa6240be832fbc8a93fdcf6023350b4a1/go/state/state_db.go#L457). Each of those entries has a considerable size (approximately 20+32 bytes for the slot ID and 32+8 for the value, assuming no overhead in the LRU cache). So each entry takes ~92 bytes. 

We could document this in the flag, or do a conversion from the flag's value to the value forwarded to the implementation. This way, the flag could be in bytes. 

But since this flag is not really expected to be set by the user, just by internal tests and tools, documenting it as an "experimental" feature and updating it to say that it is the number of cached data elements is probably the simpler and less error-prone way to go. Wdyt?

config/flags/flags.go
```diff
@@ -365,4 +365,10 @@ var (
...
+			"this to <1 will automatically set the cache capacity a hard-coded constant.",
```
HerbertJordan, line 371:
```suggestion
			"this to <1 will automatically set the cache capacity to a DB defined default value.",
```

integration/makegenesis/genesis.go
```diff
@@ -128,7 +128,7 @@ func NewGenesisBuilder() *GenesisBuilder {
 	if err != nil {
 		panic(fmt.Errorf("failed to create carmen state; %s", err))
 	}
-	carmenStateDb := carmen.CreateStateDBUsing(carmenState)
+	carmenStateDb := carmen.CreateCustomStateDBUsing(carmenState, stateDbCacheCapacity)
```
HerbertJordan, line 131: nit: I guess for this case a hard-coded small value of `1024` would have been sufficient, since the scope of creating the genesis is very small. This would have saved the need for forwarding the flag to this point.

HerbertJordan (dismissed):
Looks good, just some minor details. Thanks!

config/flags/flags.go
cabrador: I thought the unit would be incorrect. I think adding saying that its 'the number of cached data elements' is good. I will also add a suggestion that for tests value <1024 is not recommended.

HerbertJordan (approved):

PR #335, opened 2024-11-25
genesis export import test

facuMH:
This PR:
- extends the test-net to provide two new methods `GetHeaders` which returns a list of the headers of all the blocks, from 0 to latest, and `RestartWithExportImport` which stops the net, exports the genesis file, cleans the net's directory, imports the genesis, and starts the network again.
- extends `block_header_tests` to also compare hashes of original headers to the headers fetched after the restart, and prepares the test to run another iteration with the `RestartWithExportImport` once genesis file format is fixed.
- adds a test `testGenesis_NetworkCanCreateNewBlocksAfterExportImport` which also depends on genesis format being fixed. This test checks that after a `RestartWithExportImport`, the network can add more blocks and all invariants should hold between old and new blocks.

### DIFF
--- a/tests/block_header_test.go
+++ b/tests/block_header_test.go
@@ -39,19 +39,22 @@ func TestBlockHeader_SatisfiesInvariants(t *testing.T) {
 	require.NoError(err)
 	defer client.Close()
 
-	lastBlock, err := client.BlockByNumber(context.Background(), nil)
+	originalHeaders, err := net.GetHeaders()
 	require.NoError(err)
-	require.GreaterOrEqual(lastBlock.NumberU64(), uint64(numBlocks))
-
-	headers := []*types.Header{}
-	for i := int64(0); i < int64(lastBlock.NumberU64()); i++ {
-		header, err := client.HeaderByNumber(context.Background(), big.NewInt(i))
-		require.NoError(err)
-		headers = append(headers, header)
+	originalHashes := []common.Hash{}
+	for _, header := range originalHeaders {
+		originalHashes = append(originalHashes, header.Hash())
 	}
 
 	// Run twice - once before and once after a node restart.
-	for range 2 {
+	runTests := func() {
+		headers, err := net.GetHeaders()
+		require.NoError(err)
+
+		t.Run("CompareHeadersHashes", func(t *testing.T) {
+			testHeaders_CompareHeadersHashes(t, originalHashes, headers)
+		})
+
 		t.Run("BlockNumberEqualsPositionInChain", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


tests/integration_test_net.go
```diff
@@ -288,6 +288,79 @@ func (n *IntegrationTestNet) GetClient() (*ethclient.Client, error) {
...
+	fmt.Println("Network stopped. Exporting genesis file...")
+
+	// save original args
+	originalArgs := os.Args
+
+	// export
+	os.Args = []string{
+		"sonictool",
+		"--datadir", n.directory,
+		"genesis", "export", "testGenesis.g",
```
HerbertJordan, line 304: It would be cleaner to create the genesis file in a temporary directory to (a) avoid the file to be staying around after the test and (b) to avoid collisions if this function is called concurrently.

Idea: move the client's state into a sub-directory (`filepath.Join(n.directory, "state")`) and initialize the state there for regular runs. Produce the genesis file in the `n.directory` and remove and re-create the state directory inside the `n.directory`.

tests/integration_test_net.go
```diff
@@ -288,6 +288,79 @@ func (n *IntegrationTestNet) GetClient() (*ethclient.Client, error) {
...
+	}
+	err := sonictool.Run()
+	if err != nil {
+		return err
+	}
+
+	// cleant tempDir
```
HerbertJordan, line 311:
```suggestion
	// clean client state
```

tests/block_header_test.go
```diff
@@ -107,8 +110,23 @@ func TestBlockHeader_SatisfiesInvariants(t *testing.T) {
...
+	// TODO: enable when genesis is fixed
+	// require.NoError(net.restartWithExportImport())
+	// runTests()
+}
+
+func testHeaders_CompareHeadersHashes(t *testing.T, hashes []common.Hash, newHeaders []*types.Header) {
+	require := require.New(t)
 
-		require.NoError(net.Restart())
+	require.Len(newHeaders, len(hashes), "length mismatch")
```
HerbertJordan, line 127: This test may be flaky since new (empty) blocks may have been created since the restart.

It should be enough to check that `len(newHeaders) >= len(hashes)` and then just test the first `len(hashes)` headers.

HerbertJordan (commented):
Looks good, some minor feature requests though.

tests/block_header_test.go
facuMH: yes, you are right

HerbertJordan:
This changes got merged into #333

PR #336, opened 2024-11-25
Remove unnecessary effective base-fee method

HerbertJordan:
Removes the method `ftm_effectiveBaseFee` from the list of supported RPC methods. 

This method is nowhere documented and introduces complexity that restricts refactoring and other improvements. For instance, it blocks #326 due to the difficult interaction of network rules with those base-fee predictions.

Furthermore, this function's implementation seems to be wrong. For anybody to obtain the effective base fee, a simple query of the current block head would be sufficient. This head includes the current base fee.

### DIFF
--- a/ethapi/backend.go
+++ b/ethapi/backend.go
@@ -57,7 +57,6 @@ type Backend interface {
 	// General Ethereum API
 	Progress() PeerProgress
 	SuggestGasTipCap(ctx context.Context, certainty uint64) *big.Int
-	EffectiveMinGasPrice(ctx context.Context) *big.Int
 	AccountManager() *accounts.Manager
 	ExtRPCEnabled() bool
 	RPCGasCap() uint64            // global gas cap for eth_call over rpc: DoS protection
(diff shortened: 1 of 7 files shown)


ethapi/backend.go
```diff
@@ -57,7 +57,6 @@ type Backend interface {
 	// General Ethereum API
 	Progress() PeerProgress
 	SuggestGasTipCap(ctx context.Context, certainty uint64) *big.Int
-	EffectiveMinGasPrice(ctx context.Context) *big.Int
```
jenikd, line 60: Can you please also regenerate Backend mock after this change?

jenikd (commented):
Looks good. As discussed, this function was based on the `MinGasPrice` which is not used in Sonic

ethapi/backend.go
HerbertJordan: Done, thanks!

jenikd (approved):

PR #338, opened 2024-11-25
Genesis export import integration test

HerbertJordan:
Adds integration tests for the genesis export / import support.

To make this work, the export and import code was updated to also support exporting and importing blocks at non-epoch boundary height.


thaarok (approved):

thaarok:
Merged into https://github.com/Fantom-foundation/Sonic/pull/333

PR #339, opened 2024-11-25
Add integration test for JSON Genesis

HerbertJordan:
This PR adds an integration test checking the proper support for initiating the client state using a JSON genesis file.

### DIFF
--- a/integration/makefakegenesis/json.go
+++ b/integration/makefakegenesis/json.go
@@ -7,8 +7,10 @@ import (
 	"fmt"
 	"math/big"
 	"os"
+	"time"
 
 	"github.com/Fantom-foundation/go-opera/integration/makegenesis"
+	"github.com/Fantom-foundation/go-opera/inter"
 	"github.com/Fantom-foundation/go-opera/inter/drivertype"
 	"github.com/Fantom-foundation/go-opera/inter/iblockproc"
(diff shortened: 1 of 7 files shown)


thaarok (approved):
Great :+1:

integration/makefakegenesis/json.go
```diff
@@ -18,14 +23,13 @@ import (
...
-	"math/big"
-	"os"
 )
 
 type GenesisJson struct {
-	Rules    opera.Rules
-	Accounts []Account     `json:",omitempty"`
-	Txs      []Transaction `json:",omitempty"`
+	Rules         opera.Rules
+	BlockZeroTime time.Time
```
thaarok, line 30: Are you sure you want the time.Time type in the JSON?
It will be represented as a string:
```
"BlockZeroTime": "2024-11-26T10:10:33.157764548+01:00",
```

HerbertJordan: I guess this is the more human-readable format than nano-seconds. And the JSON import seems to be part of the user interface, wdyt?

I am also open to any other format. What would you suggest?

thaarok: Ok, this human readable format may be the best for this.
Please consider adding it into the example-genesis.json to provide an example of valid value there.

PR #341, opened 2024-11-26
Add check of state-root hash consistency

HerbertJordan:
This PR adds a check verifying that the state roots reported by blocks match the state roots stored in the database.

It does so by extracting witness proofs for account states on each block and verify that those proofs are rooted by a corresponding node.

While doing so, an issue regarding the initialization of archives during genesis imports has been discovered and fixed.


thaarok (approved):
Ok

jenikd (approved):

thaarok (approved):

PR #342, opened 2024-11-26
Change gasTipCap suggestion to a more meanigful estimation.

LuisPH3:
According to [Eip-1559](https://eips.ethereum.org/EIPS/eip-1559), GasTipCap (maximum priority fee) allows for transactions to be prioritized over the ones that provide a lower value for the field. 
The current estimation does not reflect the current state of the transaction pool. 

The new algorithm proposes a GasTipCap equal to the current maximum value found in the queued transactions for the pool (not yet submitted for execution). Transactions using this value will be added to the next event.


LuisPH3:
This change will not be continued.

Issue #84, opened 2024-01-24
DB: asyncflushproducer does not work correctly with flushable.SyncedPool for epoch db


thaarok:
**flushable.SyncedPool** wraps the db connection to postpone all db writes (including the db dropping) until Flush() method is called. (like db transactions)

**asyncflushproducer** wraps the db producer, to postpone any all-producer-dbs Flush until the not-flushed-size estimation exceeds some threshold. It also run the Flush in a new gorutine/thread.

The issue occurs when Opera use "lachesis-%d" database following way:
* Opera Close() and **Drop()** the "lachesis-2" database
  * SyncedPool marks the db as to be removed on the pool (db producer) Flush
* Opera re-creates the "lachesis-2" database by **OpenDB()** call
  * SyncedPool returns the old (to-be-removed) "lachesis-2" db reference (the calling code has access to the supposed-to-be-deleted data)
  * SyncedPool is still going to drop the database, it drops it on the next db Flush call (not the db-producer Flush)
* Opera **Flush()** the asyncflushproducer (typically when the Opera stops)
  * the producer has the "lachesis-2" reference, which points to the closed, dropped database, Flush causes following panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x138 pc=0x119a971]

goroutine 241 [running]:
github.com/cockroachdb/pebble.(*DB).AsyncFlush(0xc0059ea000?)
        /home/jkalina/go/pkg/mod/github.com/cockroachdb/pebble@v0.0.0-20221111210721-1bda21f14fc2/db.go:1521 +0x31
github.com/Fantom-foundation/lachesis-base/kvdb/pebble.(*Database).AsyncFlush(...)
        /home/jkalina/go/pkg/mod/github.com/!fantom-foundation/lachesis-base@v0.0.0-20240116072301-a75735c4ef00/kvdb/pebble/pebble.go:201
github.com/Fantom-foundation/lachesis-base/kvdb/pebble.(*Database).Stat(0x1?, {0x19d62e0?, 0x2a5d5e0?})
        /home/jkalina/go/pkg/mod/github.com/!fantom-foundation/lachesis-base@v0.0.0-20240116072301-a75735c4ef00/kvdb/pebble/pebble.go:326 +0x51
[...]

PR #345, opened 2024-11-26
Add integration test for rejected transactions because of insufficient balance

facuMH:
This PR solves https://github.com/Fantom-foundation/sonic-admin/issues/84
A test it added that specifically verifies that transactions are rejected if the signing account does not have enough balance to pay for `transferred value + gas_limit * max_fee_cap`. This is tested for all 4 types of transactions. 

The test uses a slice of Tx instead of a map because of the issue fixed in #346

### DIFF
--- a/tests/rejected_tx_test.go
+++ b/tests/rejected_tx_test.go
@@ -8,7 +8,6 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/ethclient"
-	"github.com/holiman/uint256"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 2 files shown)


tests/rejected_tx_test.go
```diff
@@ -0,0 +1,128 @@
...
+func TestRejectedTx(t *testing.T) {
+	require := require.New(t)
+
+	// start network
+	net, err := StartIntegrationTestNet(t.TempDir())
+	require.NoError(err)
+	defer net.Stop()
+
+	// create a client
+	client, err := net.GetClient()
```
HerbertJordan, line 23: nit: the client should to be closed.

tests/rejected_tx_test.go
```diff
@@ -0,0 +1,128 @@
...
+	require.NoError(err)
+
+	// verify receipt
+	require.Equal(receipt.Status, uint64(1))
+	require.Equal(tx.Gas(), receipt.GasUsed)
+	require.GreaterOrEqual(tx.GasFeeCap().Uint64(), receipt.EffectiveGasPrice.Uint64())
+	require.GreaterOrEqual(tx.Cost().Uint64(), receipt.EffectiveGasPrice.Uint64()*receipt.GasUsed)
+}
+
+func getChainId(t *testing.T, net *IntegrationTestNet) *big.Int {
```
HerbertJordan, line 107: nit: passing the client instead of the net would eliminate the need to create a client (that should also be closed).

tests/rejected_tx_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// create a client
+	client, err := net.GetClient()
+	require.NoError(err, "failed to get client")
+
+	chainId := getChainId(t, net)
+
+	testCases := []struct {
+		name    string
+		txMaker func(t *testing.T, account *Account, nonce uint64, price int64) *types.Transaction
```
HerbertJordan, line 30: nit: `txMaker` => `makeTx`

tests/rejected_tx_test.go
```diff
@@ -0,0 +1,128 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			maxFeeCap := getMaxFee(t, client)
+			newAccount := NewAccount()
+			nonce := getNonce(t, client, newAccount.Address())
+			testRejectedTx(t, net, newAccount.Address(), tc.txMaker(t, newAccount, nonce, maxFeeCap))
+		})
+	}
+}
+
+func testRejectedTx(t *testing.T, net *IntegrationTestNet, account common.Address, tx *types.Transaction) {
```
HerbertJordan, line 76: This test covers the gas-price part of the requirement, but the `Value` of the transaction is always zero. It would be good to run all those cases also with a Value > 0.

HerbertJordan (changes requested):
Looks good, but I think transactions with Value != 0 should also be covered.

tests/rejected_tx_test.go
```diff
@@ -0,0 +1,128 @@
...
+	require := require.New(t)
+
+	// verify estimated cost
+	estimatedCost := tx.Gas()*tx.GasFeeCap().Uint64() + tx.Value().Uint64()
+	require.Equal(tx.Cost().Uint64(), estimatedCost, "cost of transaction is not equal to balance")
```
LuisPH3, line 81:
```suggestion
	require.Equal(tx.Cost().Uint64(), estimatedCost, "transaction estimation is not equal to test estimation")
```

tests/rejected_tx_test.go
```diff
@@ -0,0 +1,128 @@
...
+			testRejectedTx(t, net, newAccount.Address(), tc.txMaker(t, newAccount, nonce, maxFeeCap))
+		})
+	}
+}
+
+func testRejectedTx(t *testing.T, net *IntegrationTestNet, account common.Address, tx *types.Transaction) {
+	require := require.New(t)
+
+	// verify estimated cost
+	estimatedCost := tx.Gas()*tx.GasFeeCap().Uint64() + tx.Value().Uint64()
```
LuisPH3, line 80: To facilitate future work, estimated costs should include blobgasfee. This should be checked independently of block processing because here we are an evaluation balance being large enough for the worst case scenario. Additionally, it may turn quite difficult to find this code (tx_pool)  when implementing the blobs deep inside the block processor, doing it now will facilitate that work. 

I will update the test for blobls, make sure that they do not charge the blobgasfee

tests/rejected_tx_test.go
```diff
@@ -0,0 +1,128 @@
...
+	chainId, err := client.ChainID(context.Background())
+	require.NoError(t, err, "failed to get chain ID::")
+	return chainId
+}
+
+func getMaxFee(t *testing.T, client *ethclient.Client) (maxFeeCap int64) {
+	block, err := client.BlockByNumber(context.Background(), nil)
+	require.NoError(t, err, "failed to get block by number")
+	baseFee := block.BaseFee().Int64()
+	maxFeeCap = int64(float64(baseFee) * 1.06)
```
LuisPH3, line 120: 1.05 should be enough, why did we change this?

LuisPH3 (changes requested):

tests/rejected_tx_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+func testRejectedTx(t *testing.T, net *IntegrationTestNet, account common.Address, tx *types.Transaction) {
+	require := require.New(t)
+
+	// verify estimated cost
+	estimatedCost := tx.Gas()*tx.GasFeeCap().Uint64() + tx.Value().Uint64()
+	if tx.Type() == types.BlobTxType {
+		estimatedCost += tx.BlobGasFeeCap().Uint64() * tx.BlobGas()
+	}
+	require.Equal(tx.Cost().Int64(), int64(estimatedCost), "transaction estimation is not equal to balance")
```
LuisPH3, line 89:
```suggestion
	require.Equal(tx.Cost().Uint64(), estimatedCost, "transaction estimation is not equal to test estimation")
```

LuisPH3 (changes requested):

HerbertJordan (approved):
Looks good, thanks!

PR #346, opened 2024-11-26
Add missing blob fields to rpc json serialization

facuMH:
To support type 3 (blob) transactions, these two missing fields need to be included in the serialization.

This missing fields triggered an error during the deserialization when fetching the last block, after a blob tx was executed.

### DIFF
--- a/tests/blob_tx_test.go
+++ b/tests/blob_tx_test.go
@@ -23,10 +23,12 @@ func TestBobTransaction(t *testing.T) {
 
 	t.Run("blob tx with empty blobs is executed", func(t *testing.T) {
 		testBlobTx_WithEmptyBlobsIsExecuted(t, ctxt)
+		lastBlockSanity(t, ctxt.client)
 	})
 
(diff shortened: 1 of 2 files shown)


tests/blob_tx_test.go
```diff
@@ -176,6 +178,15 @@ func createTestBlobTransactionWithNilSidecar(t *testing.T, ctxt *testContext) (*
 	return types.SignTx(tx, types.NewCancunSigner(chainId), ctxt.net.validator.PrivateKey)
 }
 
+func lastBlockSanity(t *testing.T, client *ethclient.Client) {
```
HerbertJordan, line 181: nit: should we check all blocks up to the last, just to make sure no new blocks are inserted between the one created with the blob transaction and the one that is tested here?

Fetching those blocks should take no time.

HerbertJordan (approved):
Nice discovery and nice fix, thanks!

tests/blob_tx_test.go
facuMH: done

HerbertJordan (approved):

PR #349, opened 2024-11-26
Add test verifying witness proof support

HerbertJordan:
This PR adds an end-to-end test for witness proofs. As a side-effect, this test also cross-checks the reported state roots in blocks.


jenikd (approved):
Maybe renaming test file and refactor this test would be fine as it is not checking only block header

PR #351, opened 2024-11-27
add `deadcode` check & delete dead code.

facuMH:
This PR removes unreachable code detected by the `deadcode` tool. 
If you think a functionality being deleted should be kept for any reason please comment on it. 

Note: the `deacode` tool runs with the flag `-test`, this means that code that is only called in tests will be considered "not deadcode". An extra analysis should be done to check which of the code only used in tests can also be deleted or is needed as infrastructure or utilities for proper testing.


debug/api.go
```diff
@@ -52,43 +48,6 @@ type HandlerT struct {
...
-// CpuProfile turns on CPU profiling for nsec seconds and writes
-// profile data to file.
-func (h *HandlerT) CpuProfile(file string, nsec uint) error {
-	if err := h.StartCPUProfile(file); err != nil {
-		return err
-	}
-	time.Sleep(time.Duration(nsec) * time.Second)
-	h.StopCPUProfile()
-	return nil
-}
```
thaarok, line 90: I confirm removing this **does not remove** possibility to use following `sonictool cli` features: :+1: 
* `debug.memStats()`
* `debug.gcStats()`
* `debug.cpuProfile('test',10)`
* `debug.goTrace('test', 10)`
* `debug.blockProfile('test', 5)`
* etc... - all this is handled by HandlerT in go-ethereum-sonic/internal/debug

tracing/tx-tracing.go
```diff
@@ -1,80 +0,0 @@
-package tracing
-
-import (
-	"sync"
-
-	"github.com/ethereum/go-ethereum/common"
-	"github.com/opentracing/opentracing-go"
```
thaarok, line 7: Please run `go mod tidy` to remove OpenTracing even from go.mod.

thaarok (approved):
I confirm removing all this dead code should be ok.

HerbertJordan (dismissed):
Looks good to me, thanks for setting this up!

If @thaarok approves the effect on the `sonictool` and `sonicd` features, I think this is good to merge.

HerbertJordan (approved):

PR #352, opened 2024-11-27
Maps block 0 and 1 to epoch 1

HerbertJordan:
This PR updates the genesis data generation to map blocks 0 and 1 to epoch 1.


jenikd (approved):
Looks good, thanks

PR #356, opened 2024-11-28
Generate different ports for each integration test net.

LuisPH3:
This change isolates each instance of an integration test by choosing a different (best-effort unused) set of ports, and turning off discovery protocol.

Although tests are currently not segregated, this PR enables the segregation of those into different processes, this change is currently waiting for the memory usage reduction.
This test enables developers to locally run multiple integration tests simultaneously.

### DIFF
--- a/tests/block_header_test.go
+++ b/tests/block_header_test.go
@@ -63,10 +63,6 @@ func testBlockHeadersOnNetwork(t *testing.T, net *IntegrationTestNet) {
 	}
 	counterAddress := receipt.ContractAddress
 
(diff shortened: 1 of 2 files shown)


HerbertJordan (approved):
Thanks for adding this feature!

cabrador (dismissed):
Nice!

LuisPH3:
After changing ports, a little issue was identified in network restart test. Test was updated to use a new client connection (using new port)

HerbertJordan (approved):
Great, thanks!

PR #359, opened 2024-11-28
Fix the block hash listed by log messages stored in the index table

HerbertJordan:
This PR fixes an issue discovered by @jenikd: when querying for log events by index-topic, the wrong block hash is reported.


jenikd (approved):
Thanks for fixing this

PR #361, opened 2024-11-28
Set correct block-hash in receipts after finishing the block

HerbertJordan:
This PR also adds an update of the block hash inside the receipt as soon as the block completes.

It does not seem like the missing block-hash set in `c_block_callbacks.go` was actually used by anybody, since all tests passed already before. However, in the spirit of defensive coding this value should be set to avoid potential future mistakes.


jenikd (approved):
Looks good, thanks

PR #362, opened 2024-11-29
Update Carmen to get proof steps in correct order

HerbertJordan:
Updates Carmen version to deliver witness proof steps in trie-navigation order.

TODO:
- [x] complete full witness proof test for updated Carmen


thaarok (approved):
Great, thanks! This is much more elegant than the interface Kamil provided before.

PR #369, opened 2024-12-02
Add a test checking the support for subscribing to log events

HerbertJordan:
This PR adds an initial test to verify that subscribing for log events through the web-socket interface is supported.

Tests that should follow:
- check that filter criteria are supported
- check that subscriptions for new head-blocks is working
- check proper isolation between multiple subscriptions
- ...


jenikd (approved):
Looks good, thanks for adding ws test

LuisPH3 (approved):

PR #370, opened 2024-12-02
Implement getAccount RPC method

cabrador:
This PR adds getAccount method to the RPC according to [this](https://www.quicknode.com/docs/ethereum/eth_getAccount) documentation.

Copy of #317 - `StorageRoot` can now be extracted from `WitnessProof`

### DIFF
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -708,6 +708,40 @@ func (s *PublicBlockChainAPI) GetBalance(ctx context.Context, address common.Add
 	return (*hexutil.U256)(state.GetBalance(address)), state.Error()
 }
 
+// GetAccountResult is result struct for GetAccount
+type GetAccountResult struct {
+	CodeHash    common.Hash    `json:"codeHash"`
+	StorageRoot common.Hash    `json:"storageRoot"`
+	Balance     *hexutil.U256  `json:"balance"`
+	Nonce       hexutil.Uint64 `json:"nonce"`
+}
+
+// GetAccount returns the information about account with given address in the state of the given block number.
+// The rpc.LatestBlockNumber and rpc.PendingBlockNumber meta block numbers are also allowed.
+// The result contains:
+// 1) CodeHash - hash of the code for the given address
+// 2) StorageRoot - storage root for the given address
+// 3) Balance - the amount of wei for the given address
+// 4) Nonce - the number of transactions for given address
+func (s *PublicBlockChainAPI) GetAccount(ctx context.Context, address common.Address, blockNrOrHash rpc.BlockNumberOrHash) (*GetAccountResult, error) {
+	state, header, err := s.b.StateAndHeaderByNumberOrHash(ctx, blockNrOrHash)
+	if err != nil {
+		return nil, err
+	}
+	proof, err := state.GetProof(address, nil)
+	if err != nil {
+		return nil, err
+	}
+	_, storageRoot, _ := proof.GetAccountElements(cc.Hash(header.Root), cc.Address(address))
+	defer state.Release()
+	return &GetAccountResult{
+		CodeHash:    state.GetCodeHash(address),
+		StorageRoot: common.Hash(storageRoot),
+		Balance:     (*hexutil.U256)(state.GetBalance(address)),
+		Nonce:       hexutil.Uint64(state.GetNonce(address)),
+	}, state.Error()
+}
+
 // AccountResult is result struct for GetProof
 type AccountResult struct {
 	Address      common.Address  `json:"address"`
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -728,17 +728,30 @@ func (s *PublicBlockChainAPI) GetAccount(ctx context.Context, address common.Add
 	if err != nil {
 		return nil, err
 	}
+	defer state.Release()
 	proof, err := state.GetProof(address, nil)
 	if err != nil {
 		return nil, err
 	}
+	codeHash, _, err := proof.GetCodeHash(cc.Hash(header.Root), cc.Address(address))
+	if err != nil {
+		return nil, err
+	}
 	_, storageRoot, _ := proof.GetAccountElements(cc.Hash(header.Root), cc.Address(address))
-	defer state.Release()
+	balance, _, err := proof.GetBalance(cc.Hash(header.Root), cc.Address(address))
+	if err != nil {
+		return nil, err
+	}
+	nonce, _, err := proof.GetNonce(cc.Hash(header.Root), cc.Address(address))
+	if err != nil {
+		return nil, err
+	}
+	u256Balance := balance.Uint256()
 	return &GetAccountResult{
-		CodeHash:    state.GetCodeHash(address),
+		CodeHash:    common.Hash(codeHash),
 		StorageRoot: common.Hash(storageRoot),
-		Balance:     (*hexutil.U256)(state.GetBalance(address)),
-		Nonce:       hexutil.Uint64(state.GetNonce(address)),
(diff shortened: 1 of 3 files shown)


tests/get_account_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	rpcClient := c.Client()
+
+	var res ethapi.GetAccountResult
+	err = rpcClient.Call(&res, "eth_getAccount", addr, rpc.LatestBlockNumber)
+	if err != nil {
+		t.Fatalf("failed to get account; %v", err)
+	}
+
+	if res.CodeHash == (common.Hash{}) {
```
HerbertJordan, line 40: It would be good if you would actually check for the real code hash here, and not just accept any non-zero value. The `rpcClient` should have functionality to ask for the code of an account if I remember correctly.

Same for the balance and nonce below.

tests/get_account_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	// Deploy the transient storage contract
+	_, deployReceipt, err := DeployContract(net, transientstorage.DeployTransientstorage)
+	if err != nil {
+		t.Fatalf("failed to deploy contract; %v", err)
+	}
+
+	addr := deployReceipt.ContractAddress
+
+	c, err := net.GetClient()
```
HerbertJordan, line 27: This client should be closed.

tests/get_account_test.go
```diff
@@ -0,0 +1,55 @@
...
+	err = rpcClient.Call(&res, "eth_getAccount", addr, rpc.LatestBlockNumber)
+	if err != nil {
+		t.Fatalf("failed to get account; %v", err)
+	}
+
+	if res.CodeHash == (common.Hash{}) {
+		t.Error("code hash should not be empty")
+	}
+
+	if res.StorageRoot == (common.Hash{}) {
```
HerbertJordan, line 44: It is possible to use the ethereum client and the `getProof` RPC function to query for an accounts storage root using a few lines of code. An extensive example also verifying proofs can be found [here](https://github.com/Fantom-foundation/Sonic/blob/1933d89b40f511df7a8481d19b8ccedd5f19e383/tests/block_header_test.go#L702-L765). A subset of those lines should be sufficient to obtain the expected storage root hash.

HerbertJordan (changes requested):
Implementation looks good to me. For the integration test, I would like to see more strict requirements for accepted results. See comments.

tests/get_account_test.go
cabrador: Since I already need to extract `WitnessProof` I just used the data for checking all of the data contained in `GetAccount`

ethapi/api.go
```diff
@@ -708,6 +708,40 @@ func (s *PublicBlockChainAPI) GetBalance(ctx context.Context, address common.Add
...
+type GetAccountResult struct {
+	CodeHash    common.Hash    `json:"codeHash"`
+	StorageRoot common.Hash    `json:"storageRoot"`
+	Balance     *hexutil.U256  `json:"balance"`
+	Nonce       hexutil.Uint64 `json:"nonce"`
+}
+
+// GetAccount returns the information about account with given address in the state of the given block number.
+// The rpc.LatestBlockNumber and rpc.PendingBlockNumber meta block numbers are also allowed.
+// The result contains:
```
kjezek, line 721: I suggest to put this part of documentation to  `GetAccountResult` and here only say that the method returns this sctruc

ethapi/api.go
```diff
@@ -708,6 +708,40 @@ func (s *PublicBlockChainAPI) GetBalance(ctx context.Context, address common.Add
...
+	if err != nil {
+		return nil, err
+	}
+	proof, err := state.GetProof(address, nil)
+	if err != nil {
+		return nil, err
+	}
+	_, storageRoot, _ := proof.GetAccountElements(cc.Hash(header.Root), cc.Address(address))
+	defer state.Release()
+	return &GetAccountResult{
```
kjezek, line 737: all the items returned here are already in the proof, i.e. it may be wasteful to fetch it again from the state.

ethapi/api_test.go
```diff
@@ -52,6 +56,61 @@ func TestGetBlockReceipts(t *testing.T) {
...
+	mockState.EXPECT().Release()
+
+	api := NewPublicBlockChainAPI(mockBackend)
+
+	account, err := api.GetAccount(context.Background(), addr, blkNr)
+	if err != nil {
+		t.Fatalf("unexpected err: %v", err)
+	}
+
+	if codeHash.Cmp(account.CodeHash) != 0 {
```
kjezek, line 93: I think hashes can be compared directly, no need to use `Cmp`.

Also I prefer using` if got, want := ....; got != want {  t.Errorf("unexpected .. , got: %s want %s", got, want) }`
It is IMHO much more readable, and reduces possible errors mixing up variables.

kjezek: btw: why below it uses the framework github.com/stretchr/testify/require
and here it is not used?

tests/get_account_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"testing"
+)
+
+func TestGetAccount(t *testing.T) {
+	net, err := StartIntegrationTestNet(t.TempDir())
+	require.NoError(t, err, "failed to start the fake network")
+	defer net.Stop()
+
+	// Deploy the transient storage contract
+	_, deployReceipt, err := DeployContract(net, transientstorage.DeployTransientstorage)
```
kjezek, line 19: I think it would be more readable to deploy `Counter` - it would make it obvious that the simplest contract was deployed. Here I had to think for a moment why transient storage is needed

kjezek (changes requested):
Thanks, looks good to me. I need some clarifications, and put some ideas for improvemtenis, please have a look

ethapi/api.go
cabrador: Yes I thought about it as well, since we already ask for witness proof we might just use the data from there, but I ended up not doing it because I simply dont like extracting everything from that. We can surely do it tho.

tests/get_account_test.go
cabrador: Sure

tests/get_account_test.go
```diff
@@ -0,0 +1,55 @@
...
+		addr,
+		nil,
+		rpc.LatestBlockNumber,
+	)
+	require.NoError(t, err, "failed call to get proof")
+
+	require.Equal(t, proofRes.CodeHash, res.CodeHash)
+	require.Equal(t, proofRes.StorageHash, res.StorageRoot)
+	require.Equal(t, proofRes.Balance, res.Balance)
+	require.Equal(t, proofRes.Nonce, res.Nonce)
```
HerbertJordan, line 54: Very nice!

HerbertJordan (approved):

ethapi/api_test.go
cabrador: Changed to use `require.Equal`

ethapi/api.go
kjezek: why did not you like it?

kjezek (approved):
I cannot think of advantage querying data, which are already available in the proof, from state, but I do not have to see all details.

ethapi/api.go
cabrador: Initialy I thought using the RPC method itself. But then I realized the data is in the proof from calling `state.GetProof()`. So I used the data from proof as you suggest.

kjezek (dismissed):

HerbertJordan (approved):

HerbertJordan (approved):

PR #371, opened 2024-12-03
Use data from proof.

cabrador:
This PR fixes RPC method `GetProof` such that it uses data from already extracted proof instead of calling the state again.


HerbertJordan (approved):
Looks good to me.

I would be curious whether extracting this information from the proof is actually faster fetching it from the DB. It might require extra time for RLP decoding and checking hashes. But these premature optimization considerations we should ignore until we actually see that this causes problems. Reading data from the proof is definitely the cleaner implementation.

PR #374, opened 2024-12-12
Fix error propagation for header functions

jenikd:
This PR fixes an error propagation when obtaining block header


HerbertJordan (approved):

HerbertJordan:
A unit test would be nice to cover this change.

PR #376, opened 2024-12-15
Caution package

facuMH:
This PR adds a utility lib that calls a function and reports whether it returned an error or not.

The function `CloseAndReportError` is meant to facilitate the check and handle of closing resources, which is one of the main unhandled errors across the code base. As we work on fixing unhandled errors (which are mostly unlikely to fail or coming from non-logical issues). 

part of https://github.com/Fantom-foundation/sonic-admin/issues/90

### DIFF
--- a/utils/caution/caution.go
+++ b/utils/caution/caution.go
@@ -1,0 +1,26 @@
+// package caution provides utility functions for handling errors and closing
+// resources.
+// This package should be only be used for sitations where errors are unexpected
+// and should be ignored or panic
+
+package caution
+
+import (
+	"fmt"
+	"io"
+)
+
+func ExecuteAndReportError(err *error, f func() error, message string) {
+	if *err == nil {
+		*err = f()
+		if *err != nil {
+			*err = fmt.Errorf("%s: %w", message, *err)
+		}
+	}
+}
+
+func CloseAndReportError(err *error, closer io.Closer, message string) {
+	ExecuteAndReportError(err, func() error {
+		return closer.Close()
+	}, message)
+}
--- a/utils/caution/caution.go
+++ b/utils/caution/caution.go
@@ -1,24 +1,67 @@
-// package caution provides utility functions for handling errors and closing
-// resources.
-// This package should be only be used for sitations where errors are unexpected
-// and should be ignored or panic
-
+// package caution provides utility functions for handling errors in defer
+// cleanup and closing resources.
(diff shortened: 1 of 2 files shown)


utils/caution/caution.go
```diff
@@ -0,0 +1,26 @@
+// package caution provides utility functions for handling errors and closing
+// resources.
+// This package should be only be used for sitations where errors are unexpected
+// and should be ignored or panic
```
LuisPH3, line 4: This comment is not correct:

`CloseAndReportError` is meant to be used in production code for every error, to facilitate error management of the pattern `defer f.Close()`. `ExecuteAndReportError` is a generalization for types that do not implement the `Closer` interface.

The interface is designed to be used in this way:
Given a code
```
func F(....) error{
    [...]
    defer f.Close()
    [...]
}
```

The new functions can be easily introduced by naming the return variable and using the symbol name:
```
func F(....) (err error) {
    [...]
    defer CloseAndReportError(&err, f, "failed to close f")
    [...]
}
```

utils/caution/caution_test.go
LuisPH3, line 1: I am missing one test case: When the function using `CloseAndReportError` already returned an error before calling defer.
This case should be handled by combining the existing value in err and the new close error. 
The error existing in err is more important than the new close error, and shall never be overwritten.

LuisPH3 (changes requested):
Unfortunately, the functions do not do what they should.  As a proof of concept they are fine, but they are not working correctly when the error is already set.

utils/caution/caution_test.go
facuMH: implementation and testing are refactor in next commit. this is handled there

utils/caution/caution.go
facuMH: documentation will be improved in next commit

LuisPH3 (approved):

thaarok (approved):
Seems ok to me.
I would personally prefer to review PR with some real-life usage instead of merging dead code into the project, but this is sufficient.

facuMH:
> Seems ok to me. I would personally prefer to review PR with some real-life usage instead of merging dead code into the project, but this is sufficient.

real-life usage is coming in the next PR, which is still a work in progress. I thought it would be easier to review if the PRs are as small and compartmentalized as possible. That said, I will keep it in mind for future work

PR #378, opened 2024-12-16
Refactor defer-close with error handling

facuMH:
This PR contributes to https://github.com/Fantom-foundation/sonic-admin/issues/90

This PR adds real life use cases of both functions provided the caution package to properly handler errors in case of `defer` and `close`. 

This PR fixes some of the unhandled errors reported by the external audit. Most (if not all) of them are from either `defer close` patterns or similar cleanups or unhandled close errors.

Test list to verify we don't have double closes (or other errors) and panic now because error is no longer ignored:
- [x] run `make test`
- [x] Norma test
    - demonet/static.yml
    - demonet/dynamic.yml
- [ ] testing suggestions ?

### DIFF
--- a/cmd/sonicd/app/launcher.go
+++ b/cmd/sonicd/app/launcher.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Fantom-foundation/go-opera/cmd/sonicd/metrics"
 	"github.com/Fantom-foundation/go-opera/config"
 	"github.com/Fantom-foundation/go-opera/config/flags"
+	"github.com/Fantom-foundation/go-opera/utils/caution"
 	"github.com/Fantom-foundation/go-opera/version"
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/console/prompt"
@@ -202,11 +203,10 @@ func initApp() {
 		return nil
 	}
 
-	app.After = func(ctx *cli.Context) error {
+	app.After = func(ctx *cli.Context) (err error) {
 		debug.Exit()
-		prompt.Stdin.Close() // Resets terminal mode.
-
-		return nil
+		caution.CloseAndReportError(&err, prompt.Stdin, "failed to reset terminal input") // Resets terminal mode.
+		return
 	}
 }
 
--- a/cmd/sonicd/app/launcher.go
+++ b/cmd/sonicd/app/launcher.go
@@ -203,10 +203,10 @@ func initApp() {
 		return nil
 	}
 
-	app.After = func(ctx *cli.Context) (err error) {
+	app.After = func(ctx *cli.Context) error {
 		debug.Exit()
-		caution.CloseAndReportError(&err, prompt.Stdin, "failed to reset terminal input") // Resets terminal mode.
-		return
+		// Close will resets terminal mode.
+		return caution.IfErrorAddContext(prompt.Stdin.Close(), "failed to reset terminal input")
 	}
 }
 
--- a/cmd/sonicd/app/launcher.go
+++ b/cmd/sonicd/app/launcher.go
@@ -203,10 +203,11 @@ func initApp() {
 		return nil
 	}
 
-	app.After = func(ctx *cli.Context) error {
+	app.After = func(ctx *cli.Context) (err error) {
 		debug.Exit()
 		// Close will resets terminal mode.
-		return caution.IfErrorAddContext(prompt.Stdin.Close(), "failed to reset terminal input")
+		caution.CloseAndReportError(&err, prompt.Stdin, "failed to reset terminal input")
+		return err
 	}
 }
 
--- a/cmd/sonictool/app/heal.go
+++ b/cmd/sonictool/app/heal.go
@@ -112,7 +112,7 @@ func healLiveFromArchive(ctx context.Context, carmenLiveDir, carmenArchiveDir st
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		defer caution.CloseAndReportError(&err, writer, "failed to close writer")
+		defer caution.CloseAndReportError(&exportErr, writer, "failed to close writer")
 		exportErr = mptio.ExportBlockFromArchive(ctx, mptio.NewLog(), carmenArchiveDir, bufWriter, uint64(recoveredBlock))
 		if exportErr == nil {
 			exportErr = bufWriter.Flush()
--- a/cmd/sonictool/app/heal.go
+++ b/cmd/sonictool/app/heal.go
@@ -17,6 +17,7 @@ import (
 	"github.com/Fantom-foundation/go-opera/cmd/sonictool/db"
 	"github.com/Fantom-foundation/go-opera/config"
 	"github.com/Fantom-foundation/go-opera/config/flags"
+	"github.com/Fantom-foundation/go-opera/utils/caution"
 	"github.com/Fantom-foundation/lachesis-base/inter/idx"
 	"github.com/ethereum/go-ethereum/log"
 	"gopkg.in/urfave/cli.v1"
@@ -93,7 +94,7 @@ func heal(ctx *cli.Context) error {
 	return nil
 }
 
-func healLiveFromArchive(ctx context.Context, carmenLiveDir, carmenArchiveDir string, recoveredBlock idx.Block) error {
+func healLiveFromArchive(ctx context.Context, carmenLiveDir, carmenArchiveDir string, recoveredBlock idx.Block) (err error) {
 	if err := os.RemoveAll(carmenLiveDir); err != nil {
 		return fmt.Errorf("failed to remove broken live state: %w", err)
 	}
@@ -102,7 +103,7 @@ func healLiveFromArchive(ctx context.Context, carmenLiveDir, carmenArchiveDir st
 	}
 
 	reader, writer := io.Pipe()
-	defer reader.Close()
+	defer caution.CloseAndReportError(&err, reader, "failed to close reader")
 	bufReader := bufio.NewReaderSize(reader, 100*1024*1024) // 100 MiB
 	bufWriter := bufio.NewWriterSize(writer, 100*1024*1024) // 100 MiB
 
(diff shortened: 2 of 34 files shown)


cmd/sonictool/app/heal.go
```diff
@@ -111,14 +112,14 @@ func healLiveFromArchive(ctx context.Context, carmenLiveDir, carmenArchiveDir st
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		defer writer.Close()
+		defer caution.CloseAndReportError(&err, writer, "failed to close writer")
```
facuMH, line 115:
```suggestion
		defer caution.CloseAndReportError(&exportErr, writer, "failed to close writer")
```

cmd/sonictool/genesis/signature.go
```diff
@@ -80,33 +82,40 @@ func CheckGenesisSignature(hash []byte, signature []byte) error {
...
 	}
 	_, err = writer.Write(signature)
 	if err != nil {
-		return err
+		err = fmt.Errorf("failed to write signature to genesis file: %w", err)
+		return
 	}
 	_, err = writer.Flush()
-	return err
+	err = fmt.Errorf("failed to flush genesis file: %w", err)
```
facuMH, line 117: this assignment needs to be inside an if, flush could return nil.

LuisPH3 (changes requested):
I saw some details that need handing;

Additionally, I would like to avoid using `caution.CloseAndReportError` for scenarios without `defer`, it does nonstandard error handling, and it may be a little convoluted to read. We can talk about a replacement.

LuisPH3 (approved):

cmd/sonictool/app/sign_genesis.go
```diff
@@ -57,18 +60,20 @@ func signGenesis(ctx *cli.Context) error {
 	return nil
 }
 
-func getGenesisHeaderHashes(genesisFile string) (ogenesis.Header, ogenesis.Hashes, error) {
+func getGenesisHeaderHashes(genesisFile string) (header ogenesis.Header, genesisHashes ogenesis.Hashes, err error) {
```
HerbertJordan, line 63: nit: Using named return values makes functions more complicated to read -- and this one in particular. Consider rewriting this into a version not using named return values.

tests/integration_test_net.go
```diff
@@ -71,7 +71,9 @@ func isPortFree(host string, port int) bool {
 	if err != nil {
 		return false
 	}
-	listener.Close()
+	if err = listener.Close(); err != nil {
+		fmt.Print("failed to close listener:", err)
```
HerbertJordan, line 75: nit: panic instead?

utils/errlock/errlock.go
```diff
@@ -5,11 +5,17 @@ import (
...
+
+	"github.com/Fantom-foundation/go-opera/utils/caution"
 )
 
 // Check if errlock is written
 func Check() error {
-	locked, reason, eLockPath, _ := read(datadir)
+	locked, reason, eLockPath, err := read(datadir)
+	if err != nil {
+		return fmt.Errorf("Node isn't allowed to start due to an error reading lock file. Please fix the issue and then delete file \"%s\". Error message:\n%w", eLockPath, err)
```
HerbertJordan, line 16: please check: The message, I guess, should be updated as follows:
```suggestion
		return fmt.Errorf("Node isn't allowed to start due to an error reading the lock file "%s". Please fix the issue. Error message:\n%w", eLockPath, err)
```

HerbertJordan (changes requested):
I do not really like that most of this refactoring is introducing named return values for non-trivial functions. It makes functions harder to read, and without an IDE highlighting it is difficult to verify that a return value name is not shadowed in a nested scope.

But I do understand that this seems to be a requirement imposed by the `caution` package, and alternatives might be even more cumbersome.

However, I would like to ask to reconsider the naming and placement of the function `IfErrorAddContext` as addressed in the comments. Please have a look.

tests/integration_test_net.go
facuMH: I was trying to handler errors without modifying the current behaviour, in that regard, a log or print seemed a better approach than panicking, which could lead to an unexpected runtime panic in a non critical situation.

HerbertJordan: This is test infrastructure - nobody will read the log unless the test fails or crashes.

cmd/sonicd/app/launcher.go
```diff
@@ -203,9 +204,8 @@ func initApp() {
 
 	app.After = func(ctx *cli.Context) error {
 		debug.Exit()
-		prompt.Stdin.Close() // Resets terminal mode.
-
-		return nil
+		// Close will resets terminal mode.
+		return caution.IfErrorAddContext(prompt.Stdin.Close(), "failed to reset terminal input")
```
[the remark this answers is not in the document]
facuMH: I will go for `AnnotateIfError`

utils/caution/caution.go
```diff
@@ -67,3 +67,11 @@ func CloseAndReportError(err *error, closer io.Closer, message string) {
 		return closer.Close()
 	}, message)
 }
+
+// IfErrorAddContext adds a message to an error, if the error is not nil.
+func IfErrorAddContext(err error, message string) error {
```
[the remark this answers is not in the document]
facuMH: moving it to it's own file in utils

cmd/sonictool/app/sign_genesis.go
facuMH: added return values to each return.

cmd/sonictool/genesis/signature.go
```diff
@@ -80,32 +83,34 @@ func CheckGenesisSignature(hash []byte, signature []byte) error {
...
+	if err = writer.Start(header, "signature", tmpDir); err != nil {
+		return fmt.Errorf("failed to write start to genesis file: %w", err)
 	}
 	_, err = writer.Write(signature)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to write signature to genesis file: %w", err)
 	}
 	_, err = writer.Flush()
+	utils.AnnotateIfError(err, "failed to flush genesis file:")
```
HerbertJordan, line 113: this should be
```suggestion
	return utils.AnnotateIfError(err, "failed to flush genesis file:")
```

Would an unhandled error produced by `AnnotateIfError` like in this case be detected by static linters?

utils/annotateError_test.go
```diff
@@ -0,0 +1,20 @@
...
+func TestAnnotateIfError_PropagatesNil(t *testing.T) {
+	if AnnotateIfError(nil, "message") != nil {
+		t.Error("AnnotateIfError should return nil when err is nil")
+	}
+}
+
+func TestAnnotateIfError_AddsContextToError(t *testing.T) {
+	err := fmt.Errorf("someError")
+	errWithContext := AnnotateIfError(err, "message")
+	require.ErrorContains(t, errWithContext, "message: someError")
```
HerbertJordan, line 19: nit: it should also check that `errors.Is(errWithContext, err)` is true.

HerbertJordan (approved):

cmd/sonictool/genesis/signature.go
facuMH: yes it does get detected.

cmd/sonicd/app/launcher.go
```diff
@@ -202,11 +203,11 @@ func initApp() {
...
 	}
 
-	app.After = func(ctx *cli.Context) error {
+	app.After = func(ctx *cli.Context) (err error) {
 		debug.Exit()
-		prompt.Stdin.Close() // Resets terminal mode.
-
-		return nil
+		// Close will resets terminal mode.
+		caution.CloseAndReportError(&err, prompt.Stdin, "failed to reset terminal input")
```
LuisPH3, line 209: this case is not defered, close can be called using normal procedures.

topicsdb/search_parallel.go
```diff
@@ -117,5 +118,7 @@ func (tt *index) scanPatternVariant(pos uint8, variant common.Hash, start uint64
 			break
 		}
 	}
-	onMatched(nil)
+	if _, err := onMatched(nil); err != nil {
+		log.Warn("searchParallel", "err", err)
```
LuisPH3, line 122: since the first time i saw this line, i dont really know what to do with it. 
The search algorithm will look for something not returning an error and producing errors may be the intended use. 

The context we add to warn is not good.
- Write a complete sentence in warn:  "error produced during parallel search"
- Add a comment explaining the issue, so that  the problem is understood in case this line turns too verbose (which could affect performance of some hot region)

@HerbertJordan @kjezek do you know what this code is about?

integration/assembly.go
```diff
@@ -81,9 +82,9 @@ func makeEngine(chaindataDir string, cfg Configs) (*abft.Lachesis, *vecmt.Index,
 	}
 	defer func() {
 		if err != nil {
```
LuisPH3, line 84:
```
//	}
func ExecuteAndReportError(err *error, f func() error, message string) {
	fErr := f()
	if fErr != nil {
		*err = errors.Join(*err, fmt.Errorf("%s: %w", message, fErr))
	}
}
```

this if is superfluous.

topicsdb/search_parallel.go
HerbertJordan: This code is the implementation of the log-filter feature. You can send RPC queries to the service asking for log messages listing a given set of topics. This function is searching for those and does a call-back on `onMatched` for each element and, as it looks like, a final one passing a `nil` to indicate that all matches have been reported and that there are no more.

The only time this function is actually used is in line 94 above, where `onMatched` returns an error if, for instance, the execution context is canceled. Another option is that the consumer of the match failed to consume it, which seems to be ignored, as shown in line 116.

I guess the best option here, matching the action in line 116, is to simply ignore the error.

HerbertJordan (approved):

LuisPH3:
Please, Add one deferred error missing : `gossip/handler.go:700:19:       defer msg.Discard()`

facuMH:
moved to https://github.com/0xsoniclabs/sonic/pull/7

PR #380, opened 2024-12-18
Handler errors in cmd, debug and more

facuMH:
This PR is part of https://github.com/Fantom-foundation/sonic-admin/issues/90

It contributes by handling errors that do not need the `caution` util, nor occur in tests. These could have a real production impact, so they should be thoroughly tested much like #378

- [x] run `make test`
- [x] Norma test
    - [x] demonet/static.yml
    - [x] demonet/dynamic.yml
- [ ] testing suggestions ?


LuisPH3:
migrated  https://github.com/0xsoniclabs/sonic/pull/7

PR #383, opened 2024-12-21
Add verification support to rule update code

HerbertJordan:
This PR introduces a network rule verification infrastructure and adds support for validating rule changes on the network.

The rule verification code covers basic rule properties by constraining the domain of various rule parameters. The main goal of those constraints is to prevent an accidental network stall. These validation steps are to be considered a last line of defense to safe-guard the network lifelines. However, they are necessarily incomplete. Any change of network rules should be tested extensively before being applied on production networks.

To facilitate a safe roll-out, the validation step is ignored by default. Only a rule-update will enable this features. To prevent a network split, the validation features should only be updated once sufficient (ideally all) nodes on the network have been updated to support this feature.

### DIFF
--- a/gossip/blockproc/drivermodule/driver_txs.go
+++ b/gossip/blockproc/drivermodule/driver_txs.go
@@ -23,7 +23,8 @@ import (
 )
 
 const (
-	maxAdvanceEpochs = 1 << 16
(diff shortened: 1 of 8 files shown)


thaarok (approved):

HerbertJordan:
This PR will be re-opened in the new repository ([here](https://github.com/0xsoniclabs/sonic)).

PR #390, opened 2025-01-14
Get current version from git tag 

facuMH:
This PR adds a mechanism that uses the git tag (if any) as release version. Hence preventing the task to update both the version file as well as the release tag in git. It also should check if there have been any commits since the last label, and if the build is a dirty build (aka has uncommitted changes)

the git tag is "v2.0.1" => Major=2, Minor=0, Patch=1, Meta=""
the git tag is "v2.0.1-rc1" => Major=2, Minor=0, Patch=1, Meta="rc1"
the git tag is "v2.0.1" but there are extra commits on top: => Major=2, Minor=0, Patch=1, Meta="dev"
the git tag is "v2.0.1-rc1" with extra commits on top => Major=2, Minor=0, Patch=1, Meta="rc1-dev"
there is no git tag (e.g. development) => handle like "v0.0.0" with extra commits on top

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ sonicd:
 	go build \
 			-ldflags "-s -w -X github.com/Fantom-foundation/go-opera/config.GitCommit=$${GIT_COMMIT} \
 							-X github.com/Fantom-foundation/go-opera/config.GitDate=$${GIT_DATE} \
-							-X github.com/Fantom-foundation/go-opera/version.GitTag=$${GIT_TAG}" \
+							-X github.com/Fantom-foundation/go-opera/version.Version=$${GIT_TAG}" \
             -o build/sonicd \
             ./cmd/sonicd && \
 			./build/sonicd version
@@ -24,7 +24,7 @@ sonictool:
 	go build \
 			-ldflags "-s -w -X github.com/Fantom-foundation/go-opera/config.GitCommit=$${GIT_COMMIT} \
 							-X github.com/Fantom-foundation/go-opera/config.GitDate=$${GIT_DATE} \
-							-X github.com/Fantom-foundation/go-opera/version.GitTag=$${GIT_TAG}" \
+							-X github.com/Fantom-foundation/go-opera/version.Version=$${GIT_TAG}" \
             -o build/sonictool \
             ./cmd/sonictool && \
 			./build/sonictool --version
--- a/Makefile
+++ b/Makefile
@@ -6,20 +6,28 @@ GOPROXY ?= "https://proxy.golang.org,direct"
 sonicd:
 	GIT_COMMIT=`git rev-list -1 HEAD 2>/dev/null || echo ""` && \
 	GIT_DATE=`git log -1 --date=short --pretty=format:%ct 2>/dev/null || echo ""` && \
+	GIT_TAG=`echo $(call get_git_tag)` && \
 	GOPROXY=$(GOPROXY) \
 	go build \
-	    -ldflags "-s -w -X github.com/Fantom-foundation/go-opera/config.GitCommit=$${GIT_COMMIT} -X github.com/Fantom-foundation/go-opera/config.GitDate=$${GIT_DATE}" \
-	    -o build/sonicd \
-	    ./cmd/sonicd
+			-ldflags "-s -w -X github.com/Fantom-foundation/go-opera/config.GitCommit=$${GIT_COMMIT} \
+							-X github.com/Fantom-foundation/go-opera/config.GitDate=$${GIT_DATE} \
+							-X github.com/Fantom-foundation/go-opera/version.GitTag=$${GIT_TAG}" \
+            -o build/sonicd \
+            ./cmd/sonicd && \
+			./build/sonicd version
 
 sonictool:
 	GIT_COMMIT=`git rev-list -1 HEAD 2>/dev/null || echo ""` && \
 	GIT_DATE=`git log -1 --date=short --pretty=format:%ct 2>/dev/null || echo ""` && \
+	GIT_TAG=`echo $(call get_git_tag)` && \
 	GOPROXY=$(GOPROXY) \
 	go build \
-	    -ldflags "-s -w -X github.com/Fantom-foundation/go-opera/config.GitCommit=$${GIT_COMMIT} -X github.com/Fantom-foundation/go-opera/config.GitDate=$${GIT_DATE}" \
-	    -o build/sonictool \
-	    ./cmd/sonictool
+			-ldflags "-s -w -X github.com/Fantom-foundation/go-opera/config.GitCommit=$${GIT_COMMIT} \
+							-X github.com/Fantom-foundation/go-opera/config.GitDate=$${GIT_DATE} \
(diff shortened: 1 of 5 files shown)


version/version_test.go
```diff
@@ -39,3 +39,38 @@ func TestAsBigInt(t *testing.T) {
...
+	}{
+		"v1.2.3":                       {major: 1, minor: 2, patch: 3},
+		"v1.2.3-alpha":                 {major: 1, minor: 2, patch: 3, meta: "alpha"},
+		"v1.2.3-alpha-dirty":           {major: 1, minor: 2, patch: 3, meta: "alpha-dirty"},
+		"some-non.stan-dard.12tag":     {},
+		"!`@#$%^&*()_{}|:<>?[]\\;',./": {},
+		"myTestTag":                    {},
+	}
+
+	originalGitTag := GitTag
```
HerbertJordan, line 60: This should no longer be needed.

version/version.go
```diff
@@ -2,46 +2,84 @@ package version
...
+	"strings"
 )
 
-const (
-	VersionMajor = 2  // Major version component of the current release
-	VersionMinor = 0  // Minor version component of the current release
-	VersionPatch = 1  // Patch version component of the current release
-	VersionMeta  = "" // Version metadata to append to the version string
+var (
+	GitTag       = "" // Git tag of this release
```
HerbertJordan, line 10: Please make this a const.

version/version_test.go
```diff
@@ -39,3 +39,38 @@ func TestAsBigInt(t *testing.T) {
...
+	for tag, want := range tests {
+		versionMajor = 0
+		versionMinor = 0
+		versionPatch = 0
+		versionMeta = ""
+
+		versionMajor, versionMinor, versionPatch, versionMeta = parseVersion(tag)
```
HerbertJordan, line 67: Please do not use the variables for this test.

```suggestion
		versionMajor, versionMinor, versionPatch, versionMeta := parseVersion(tag)
```

version/version.go
```diff
@@ -2,46 +2,84 @@ package version
...
+	versionMajor = 0  // Major version component of the current release
+	versionMinor = 0  // Minor version component of the current release
+	versionPatch = 0  // Patch version component of the current release
+	versionMeta  = "" // Meta information of the current release
```
HerbertJordan, line 14: This is not nowhere used, so I guess we should remove it.

version/version.go
```diff
@@ -2,46 +2,84 @@ package version
...
 )
 
 // Version holds the textual version string.
 var Version = func() string {
```
HerbertJordan, line 18: `Version` could be a constant by implementing this as follows:
```
// Version holds the textual version string.
const Version = GitTag

func init() {
	// in case of no tag or small/irregular tag, return it as is
	if len(GitTag) >= 2 {
		versionMajor, versionMinor, versionPatch, versionMeta = parseVersion(GitTag)
	}
}
```

In this case, `GitTag` could be removed and the make file could just set the `Version` directly. Wdyt?

HerbertJordan (changes requested):
Looks good. However, I added some minor code changes to reduce mutable globals. Please have a look.

Makefile
```diff
@@ -73,4 +81,22 @@ deadcode:
...
 .PHONY: lint
-lint: vet staticcheck deadcode # errorcheck
\ No newline at end of file
+lint: vet staticcheck deadcode # errorcheck
+
+# get_git_tag get the last git tag and append -dev if there are commits since
+# the last tag and -dirty if there are uncommitted changes
+define get_git_tag
+    $(shell \
+        GIT_TAG=$$(git describe --tags --abbrev=0 2>/dev/null); \
```
LuisPH3, line 90: There is a thing we need to test here, but I guess we will only detect it after the project moves forward. 

For previous tags to be available at this point, we assume that the repository has been cloned with the complete story. This may not be the case if the repo is cloned with "--depth 1"  to speed up downloads, or if the code is downloaded using a tarball. 

I would discourage the use of TAG+commit count, because of the bloat on download time that this produces on each CI run.  For each PR, for each development cycle.
Additionally, no one should care if this is one, two or n commits after a tag, if it's not tagged, the commit should not be used for release testing nor deployment. IMO this kind of tagging only produces more problems than benefits. 

Please do not confuse this with the dirty flag, which is of maximum importance. 

Nevertheless, if this is what we want to have... I guess this is it. 
We need to check downstream projects using "clone --depth 1" and see how does this script behave.  Worst case scenario, this script cashes during build. Same for tarballs.

LuisPH3: FYI @HerbertJordan

facuMH: ran `git clone --branch=facundo/tag_to_version --depth=1 git@github.com:Fantom-foundation/Sonic.git` build did not fail running version commands also did ran fine reported version was empty as expected,  and only commit is reported for sonictool.

LuisPH3: There is a second problem with the count after tags, sorry that I did not remember yesterday.
Tag+count is not unique: any commit can be the shared between branches and each commit on each branch would be named the same regarding the common ancestor; example:

The first commit in release X.Y branch is still part of the main branch, this commit will be tagged RC1,and therefore there will be two RC1-1, one in the release branch and the other in the main branch. 

I think this naming brings more problems than benefits.

version/version.go
facuMH: I like the idea of removing the 4 global variables, and just having them as return from `parseVersion`. So I will keep only `Version` as a package variable, which gets assigned at linktime. `GitTag` is replaced by `Version`.

version/version.go
```diff
@@ -64,21 +49,25 @@ func parseVersionComponent(parts []string, index int, stripPrefix bool) int {
...
-	if (strings.Split(GitTag, "-")[0] != "") && (gitDate != "") {
+	if (strings.Split(Version, "-")[0] != "") && (gitDate != "") {
 		vsn += "-" + gitDate
 	}
 	return vsn
 }
 
 func AsString() string {
+	// meta is not used for now, so we ignore it.
+	versionMajor, versionMinor, versionPatch, _ := parseVersion(Version)
```
HerbertJordan, line 64: By moving the version number parsing from the init code to a lazy-evaluated place, you effectively made the final `sonicd version` command in the Makefile for checking whether the version string could be correctly parsed useless -- since it is just printing the `Version` constant.

facuMH: even if `parseVersion` is called in init, it just initializes the 3 version variables, which are only used when comparing version with the network. Both `sonicd` and `sonictool` when running version cmd, they just return the `Version`. I can change it back to have `parseVersion` in the init, but it does not affect the version cmds

LuisPH3 (dismissed):

LuisPH3 (approved):

version/version.go
HerbertJordan: If it is in the `init` it does affect the version command in the sense that it gets executed. And if there is a parsing error in the version number that causes the parser to panic because of a very miss-shaped version tag, then we would detect it immediately at the end of the build.

The way it is now, we would only see it crashing if `AsString()` is called, which basically never happens. Would be bad if this would happen the first time in production.

HerbertJordan (commented):
This PR still needs a fix making sure that the version parsing of a release build is detected early -- not just while running in production.

PR #391, opened 2025-01-22
RPC gas capping

jenikd:
This PR fixes gas cap of the RPC requests. Because of [Tosca](https://github.com/Fantom-foundation/Tosca/blob/4da02eca5a6c16bfcec36c1c401957cab71b8e57/go/tosca/interpreter.go#L137) maximum gas has to be limited to MaxInt64. If it was higher, the EVM immediately failed because of out of gas.


HerbertJordan (approved):
Looks good, thanks for fixing this!

PR #393, opened 2025-01-28
Fix recognition of intrinsic gas error

jenikd:
This PR fixes situation, when gas estimation failed because of intrinsic gas too low error. Error was not correctly recognized, because raised error was from go-ethereum libary, not the one from sonic repository.

This is a copy of https://github.com/0xsoniclabs/sonic/pull/4 because we want this include into 2.0.2


HerbertJordan:
@wsodsong is there a plan to have another release candidate for `v2.0.2` to include this or should this be targeted for `v2.0.3`?

HerbertJordan:
Closing this change since it will not be part of `v2.0.2`. Version `v2.0.3` will be derived from the new repository.

PR #403, opened 2025-04-01
Check archive block height before sending subscribers notification

jenikd:
This PR solves issue, when there is a asynchronous write to the archive state database. This approach was added not to block block processing with write to archive as this is expensive operation.
In some cases update was done only to live database and RPC subscribers were notified about new block, but it was not yet written to archive.

### DIFF
--- a/gossip/mps_test.go
+++ b/gossip/mps_test.go
@@ -575,18 +575,23 @@ func TestMisbehaviourProofsWrongBlockEpoch(t *testing.T) {
 	require.ErrorIs(err, heavycheck.ErrUnknownEpochBVs)
 
 	goodEpochMp := copyMP(correctMp)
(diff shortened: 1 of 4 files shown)


gossip/mps_test.go
HerbertJordan, line 1: I am not familiar with this part of the code, so I can not judge the implications of these changes. Maybe @thaarok can confirm that these changes are as expected or that this feature is no longer used?

HerbertJordan (approved):

gossip/mps_test.go
thaarok: this feature is no longer used, so it should be ok

jenikd (commented):
Thanks for feedback

thaarok (approved):

PR #408, opened 2025-09-09
Check NoArchiveError when sending block notifications

jenikd:
This PR fixes situation, when there is no archive database for checking current block height when sending subscribers block notification
Related to 
https://github.com/0xsoniclabs/sonic-admin/issues/175
https://github.com/0xsoniclabs/sonic/pull/214


gossip/service_test.go
HerbertJordan, line 1: I am assuming this is a back-port of the tests in the Sonic repo, since they look like a 1:1 copy; Please let me know if there are some changes to have an eye on.

HerbertJordan (approved):
Thanks for back-porting this fix.

gossip/service_test.go
jenikd: Yes, this is a copy of the fix in Sonic repo.

